PR #412, opened 2023-11-21
fix(overlays): use getComputedStyle over computedStyleMap

rdunk:
Replaces the use of the experimental `computedStyleMap` with `window.getComputedStyle` to get an element's `display` property. The former method is not supported by Firefox.


mariuslundgard (approved):
🙏

PR #416, opened 2023-11-23
fix(react-loader): Fix useQuery - snapshot state is now updated when initial changes

### LINKED ISSUE #415: useQuery data is stale after same route navigation in Remix
**Describe the bug**

The data returned from the new useQuery hook (when used in Remix) is stale after same route navigation.

**To Reproduce**

In the Remix example in this repo, if you add a Link component in the shoes.$slug route, that takes you from a shoe to another, the url changes but the data in the page stays the same as the previous shoe.

It does work for normal navigations using an a tag but it doesn't work when using client navigation use Remix's Link component. 

**Expected behavior**

The data returned from the useQuery hook should be updated after each navigation even if both pages use same Remix route.

thomasKn:
Fixes #415

### DIFF
--- a/packages/react-loader/src/createQueryStore.ts
+++ b/packages/react-loader/src/createQueryStore.ts
@@ -29,11 +29,22 @@ export const createQueryStore = (
     params: QueryParams = DEFAULT_PARAMS,
     options: UseQueryOptions<QueryResponseResult> = {},
   ) => {
-    const [initial] = useState(() =>
+    const [initial, setInitial] = useState(() =>
       options.initial
         ? { perspective: 'published' as const, ...options.initial }
         : undefined,
     )
+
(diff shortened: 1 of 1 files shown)


thomasKn:
@stipsan you can test the issue [here](https://headless-clone.vercel.app/) by switching between `home` and `about-us` routes. The issue only happens outside of the studio.
Also here is a [link to our repo](https://github.com/buildheadless/headless/tree/main/templates/hydrogen-theme) if needed.

stipsan:
Thanks @thomasKn! Reviewing your PR now :)

thomasKn:
> Thanks @thomasKn! Reviewing your PR now :)

It works perfectly thanks @stipsan!

PR #553, opened 2023-12-12
fix: support `aria-label` on `<svg>` elements

stipsan:


rdunk:
Closing in favour of #1161

PR #581, opened 2023-12-26
fix: allow opting out of `cache: no-store` fetch option in `@sanity/preview-url-secret`

evankirkiles:
It is currently not possible to use the `validatePreviewUrl` function from `@sanity/preview-url-secret` in a CloudFlare Worker, as the Worker runtime's `fetch` implementation (annoyingly) throws in the presence of the `cache` request init keyword instead of failing silently or simply providing an implementation stub.

Ideally, CloudFlare updates the Worker runtime to not throw in this case, as the `cache` option is part of the fetch standard. Until that happens, this PR allows users to opt out of setting the `cache: "no-store"` option in `validateSecret`'s request init object, allowing `@sanity/preview-url-secret` to be used in CloudFlare Workers.

Related issues:
-  https://github.com/cloudflare/workerd/issues/698

### DIFF
--- a/packages/preview-url-secret/src/validateSecret.ts
+++ b/packages/preview-url-secret/src/validateSecret.ts
@@ -25,7 +25,7 @@ export async function validateSecret(
     {
       tag,
       // @ts-expect-error -- the `cache` option is valid, but not in the types when NextJS typings aren't installed
-      ...(!disableCacheStore ? { cache: 'no-store'} : undefined)
+      ...(!disableCacheNoStore ? { cache: 'no-store'} : undefined)
     },
   )
   if (!result?._id || !result?._updatedAt || !result?.secret) {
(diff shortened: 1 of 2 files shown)


evankirkiles:
Alternatively, we could just allow overriding the request init fields (with the default value being `cache: "no-store"` to not break existing behaviors).

packages/preview-url-secret/src/validateSecret.ts
```diff
@@ -24,7 +25,7 @@ export async function validateSecret(
     {
       tag,
       // @ts-expect-error -- the `cache` option is valid, but not in the types when NextJS typings aren't installed
-      cache: 'no-store',
+      ...(!disableCacheStore ? { cache: 'no-store'} : undefined)
```
stipsan, line 28:
```suggestion
      ...(!disableCacheNoStore ? { cache: 'no-store'} : undefined)
```

stipsan (approved):
Thanks! 🙌

PR #639, opened 2024-01-08
feat: add share URL button and improve popup window mode

stipsan:
# New experimental share URL button

Opt in to using it by:

```tsx
presentationTool({
  unstable_showUnsafeShareUrl: true, // defaults to `false`
})
```

Once enabled there's now a share menu next to the location input, rather than inside it, that opens a popover:

image

# Improved popup mode

When using the "Open preview" popup mode it's now live, just like the iframe preview:

https://github.com/sanity-io/visual-editing/assets/81981/46464e73-7126-4159-8ed5-c6e2bd1c7a5f

### DIFF
--- a/apps/next/src/app/shoes/VisualEditing.tsx
+++ b/apps/next/src/app/shoes/VisualEditing.tsx
@@ -55,7 +55,11 @@ export default function VisualEditing() {
 
   useLiveMode({ client: stegaClient })
   useEffect(() => {
-    if (process.env.NEXT_PUBLIC_VERCEL_ENV !== 'preview' && window === parent) {
+    if (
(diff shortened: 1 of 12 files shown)


rdunk:
I've added a small fix for the `<ShareUrlMenu>` not hiding when clicking outside of the element. It seems if the `<Menu>` was inside the component function, it would prevent default behaviour. I'm not sure why?

But this is a great feature! LGTM!

rdunk (approved):
I've added a small fix for the `<ShareUrlMenu>` not hiding when clicking outside of the element. It seems if `<Menu>` was inside the component function, it would prevent default behaviour. I'm not sure why?

But this is a great feature! LGTM!

PR #644, opened 2024-01-08
feat(svelte-loader): simplify api, expose shortcut setup

rdunk:
This adds the API simplifications and improvements present in the React loader to the Svelte loader. Largely follows the same pattern and file structure, e.g. `defineUseQuery`.


stipsan (approved):
Great stuff!!!

PR #686, opened 2024-01-13
fix(presentation): allow forward navigation

rdunk:
This is a rework of parameter and router state handling to fix navigation.

Using `setParams` and `params` as a source of truth was causing unwanted history navigations, meaning history forward functionality was broken. After triggering back in history (i.e. browser back button), the tool would push to the history stack.

Would be wise to test more before merging as this significantly alters the state handling!


stipsan (approved):
Looks good, lets go!

PR #762, opened 2024-01-26
feat: add option to use cdn in loadquery

SimeonGriggs:
In my application, I have a need to selectively set `useCdn` based on factors other than those that impact the current perspective or stega settings. So exposing this option when invoking `loadQuery` is required.


stipsan (approved):
Thank you! 💖

PR #780, opened 2024-01-27
feat(svelte-loader): preview support

rdunk:
This adds an (optional) opinionated preview implementation to the Svelte loader package.

To support that, the PR also updates the cache fetch definition in core loader to support explicitly passing `perspective` and `useCdn` options, [see diff here](https://github.com/sanity-io/visual-editing/pull/780/files#diff-849fcb9a221c95f0a14de41f029e332116f7a63fad66a15c77968b59d80c475dL115-R129). Not sure if that has any implications for other loaders?


apps/svelte/src/lib/server/sanity.ts
```diff
@@ -0,0 +1,10 @@
+import { SANITY_API_READ_TOKEN } from '$env/static/private'
+import { client } from '$lib/sanity'
+
+export const serverClient = client.withConfig({
+  token: SANITY_API_READ_TOKEN,
+  stega: {
+    ...client.config().stega,
+    enabled: true,
+  },
```
stipsan, line 9: You can use the short-hand for toggling stega:
```suggestion
  stega: true,
```

packages/svelte-loader/README.md
```diff
@@ -55,33 +56,112 @@ export const client = createClient({
...
+
+```ts
+// src/lib/server/sanity.ts
+import { SANITY_API_READ_TOKEN } from '$env/static/private'
+import { client } from '$lib/sanity'
+
+export const serverClient = client.withConfig({
+  token: SANITY_API_READ_TOKEN,
+  // Optionally enable stega
   // stega: {
```
stipsan, line 75:
```suggestion
  // stega: true
```

packages/svelte-loader/README.md
```diff
@@ -55,33 +56,112 @@ export const client = createClient({
...
-  //   studioUrl: PUBLIC_SANITY_STUDIO_URL,
+  //   ...client.config().stega,
+  //   enabled: true,
```
stipsan, line 77:
```suggestion
```

packages/svelte-loader/README.md
```diff
@@ -55,33 +56,112 @@ export const client = createClient({
...
   // },
```
stipsan, line 78:
```suggestion
```

stipsan (dismissed):
>To support that, the PR also updates the cache fetch definition in core loader to support explicitly passing perspective and useCdn options, [see diff here](https://github.com/sanity-io/visual-editing/pull/780/files#diff-849fcb9a221c95f0a14de41f029e332116f7a63fad66a15c77968b59d80c475dL115-R129). Not sure if that has any implications for other loaders?

No implications beyond giving them the option to use it as well 😄

PR #803, opened 2024-01-31
await loadQuery

abteilung:
const initial should be awaited.


rdunk:
Thanks for the PR, this change was incorporated into the README updates in https://github.com/sanity-io/visual-editing/pull/780 which is now merged.

PR #898, opened 2024-02-13
refactor(visual-editing): use state over ref for channel instance

rdunk:
Fixes failing overlays: Channel was stored as a ref, which means no reactivity, which is needed in order to render the Overlays element.

### DIFF
--- a/packages/visual-editing/src/ui/useChannel.tsx
+++ b/packages/visual-editing/src/ui/useChannel.tsx
@@ -1,6 +1,6 @@
 import { type ChannelsNode, createChannelsNode } from '@sanity/channels'
 import type { VisualEditingConnectionIds } from '@sanity/visual-editing-helpers'
-import { useEffect, useRef } from 'react'
+import { useEffect, useState } from 'react'
 
 import {
   VisualEditingChannelReceives as Receives,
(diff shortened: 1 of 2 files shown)


packages/visual-editing/src/ui/useChannel.tsx
```diff
@@ -12,19 +12,19 @@ import {
...
       id: 'overlays' satisfies VisualEditingConnectionIds,
       connectTo: 'presentation' satisfies VisualEditingConnectionIds,
     })
-    channelRef.current = channel
+    setChannel(channelInstance)
     return () => {
-      channel.destroy()
-      channelRef.current = undefined
+      channelInstance.destroy()
+      setChannel(undefined)
```
stipsan, line 25: I don't think it's necessary to call `setChannel(undefined)` here
```suggestion
```
as the `useChannel` hook is used during render, and React isn't going to call teardown of this useEffect and then call another render.

It's always going to be either:
```
render <VisualEditing /> -> calls useChannel()
run useChannel -> useEffect setup
// if strict mode
run useChannel -> useEffect teardown
run useChannel -> useEffect setup
render <VisualEditing /> -> calls useChannel()
run useChannel -> useEffect, deps is [] and so it's skipped
```
Or, in the event of unmounting (`enableVisualEditing` calls `root.unmount()`):
```
// Initial render (not strict mode)
render <VisualEditing /> -> calls useChannel()
run useChannel -> useEffect setup
// root.unmount()
run useChannel -> useEffect teardown
// instead of doing a last render of <VisualEditing />, its related DOM nodes are removed and `useRef` state is freed up
```

stipsan (dismissed):

stipsan (approved):

packages/visual-editing/src/ui/useChannel.tsx
rdunk: Sorry @stipsan, just saw I made the change with no response to your comment (I feel I didn't hit submit), hope that didn't seem passive aggressive. 😆

PR #913, opened 2024-02-14
chore: support presentation tools with custom names

stipsan:
Prefixing with `chore` instead of `fix` since the comment intent feature isn't released yet.

`mode` should always use `presentation`, it can't use the custom name.


rdunk (approved):
LGTM

PR #931, opened 2024-02-16
doc: correct Next.js draftMode().isEnabled usage

rvanlaarhoven:
This updates the Next.js app router example to the correct way of using `draftMode().isEnabled`.


stipsan (approved):
Thank you!

PR #942, opened 2024-02-17
feat(visual-editing): report documents on page

rdunk:
Adds document reporting from the `visual-editing` package. This means loaders or preview kit are not strictly required for reporting documents.

Each method of document reporting now provides a key to `setDocumentsOnPage`, so the cache can store them as separate objects, which are merged and deduped by when returned as `documentsOnPage`. The entire cache is replaced when some new documents are received and the frame url has changed.

Also removes some unnecessary JSON stringifying and parsing when traversing and decoding overlay nodes.


stipsan (approved):
Great stuff! 👍✨

PR #981, opened 2024-02-22
feat: add refresh mutation API

stipsan:
Docs will be updated after the API is available in a stable release of `sanity`


packages/presentation/src/loader/RevalidateTags.tsx
```diff
@@ -1,59 +0,0 @@
...
-            console.warn('Type is missing, skipping revalidation', update)
-            return
-          }
-          const slug: string | undefined = update.result?.slug?.current
-            ? `${type}:${update.result.slug.current}`
-            : undefined
-          const tags = [getPublishedId(update.documentId), type, slug!].filter(
-            Boolean,
-          ) as [string, string, string]
-          channel.send('loaders', 'loader/revalidate-tags', {
```
stipsan, line 48: I forgot to remove this event, making this note so I remember tomorrow

PR #982, opened 2024-02-22
fix: allow encoding draft ids

rdunk:


packages/presentation/src/loader/LoaderQueries.tsx
```diff
@@ -430,10 +430,7 @@ export function turboChargeResultIfSourceMap<T = unknown>(
         return undefined
       }
       // If there's a displayed document, always prefer it
-      if (
-        liveDocument?._id &&
-        getPublishedId(liveDocument._id) === getPublishedId(sourceDocument._id)
-      ) {
+      if (liveDocument?._id === sourceDocument._id) {
```
rdunk, line 433: This seems to fix the mismatch we discussed, but I assume there was a a good reason for comparing as published IDs? Will this change break something else?

packages/core-loader/src/live-mode/enableLiveMode.ts
rdunk, line 1: There seemed to be an issue with (at least) the Svelte loader where there was a mismatch between sourceMap documents and perspective. Setting the entire object rather than by individual keys seems to resolve it, still trying to work out why. 🤷

packages/presentation/src/loader/LoaderQueries.tsx
```diff
@@ -53,6 +53,8 @@ export default function LoaderQueries(props: LoaderQueriesProps): JSX.Element {
   useEffect(() => {
     if (channel) {
       const { projectId, dataset } = clientConfig
+      // @todo - Can this be migrated/deprecated in favour of emitting
+      // `presentation/perspective` at a higher level?
```
stipsan, line 57: Sure, but we would still need to dispatch it to two channels anyway, right?

stipsan (approved):

packages/presentation/src/loader/LoaderQueries.tsx
rdunk: Yes exactly, and I think we can just pass an array of channel IDs to `send`.

Would need to listen for this event on the other side over `loader/perspective` so left in place to avoid a breaking change for now, but I suppose duplicating wouldn't be a problem.

stipsan (approved):

PR #1060, opened 2024-03-06
fix(presentation): prevent horizontal scroll on narrow document pane

rdunk:
Finally, adds support for collapsed state and prevents overscroll.

Collapsed state is triggered at 640px, this could be adjusted.

https://github.com/sanity-io/visual-editing/assets/8467307/96919201-108f-4685-9d7f-9ee971a5bea3

### DIFF
--- a/packages/presentation/src/editor/DocumentPane.tsx
+++ b/packages/presentation/src/editor/DocumentPane.tsx
@@ -64,15 +64,13 @@ export function DocumentPane(props: {
     setErrorParams(null)
   }, [documentId, documentType, params])
 
-  const { setLayoutCollapsed } = useStructureTool()
-  const handleRootCollapse = useCallback(
-    () => setLayoutCollapsed(true),
-    [setLayoutCollapsed],
-  )
-  const handleRootExpand = useCallback(
-    () => setLayoutCollapsed(false),
-    [setLayoutCollapsed],
-  )
+  const { layoutCollapsed, setLayoutCollapsed } = useStructureTool()
+  const handleRootCollapse = useCallback(() => {
+    setLayoutCollapsed(true)
+  }, [setLayoutCollapsed])
+  const handleRootExpand = useCallback(() => {
+    setLayoutCollapsed(false)
+  }, [setLayoutCollapsed])
 
   if (errorParams) {
     return (
@@ -99,9 +97,10 @@ export function DocumentPane(props: {
   return (
     <ErrorBoundary onCatch={setErrorParams}>
       <PaneLayout
-        style={{ height: '100%' }}
+        style={layoutCollapsed ? { minHeight: '100%' } : { height: '100%' }}
         onExpand={handleRootExpand}
         onCollapse={handleRootCollapse}
+        minWidth={640}
       >
         <PresentationPaneRouterProvider
           onDeskParams={onDeskParams}
--- a/packages/presentation/src/panels/Panel.tsx
+++ b/packages/presentation/src/panels/Panel.tsx
@@ -17,7 +17,7 @@ interface PanelProps extends PropsWithChildren {
 }
 
 const Root = styled.div`
-  overflow: hidden;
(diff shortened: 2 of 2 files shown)


stipsan (commented):
There's odd behavior when you click "Inspect document" when the sidebar is below the breakpoint:
image

And the "Compare" button disappears, it should look like this:
image
But instead it's just gone:
image

rdunk:
Thanks for noticing these, both features of the document pane I guess.

Compare button: Disappears in Structure tool when collapsed too, is this a bug?
Inspect Document: I thought this was actually nicer, seems easier to inspect when you have a narrow pane.

Could potentially workaround and prevent the collapsed state unless inspecting comments or AI assist? Would that be preferable?

stipsan:
> Compare button: Disappears in Structure tool when collapsed too, is this a bug?
Seems like it, let's notify Studio DX?

> Inspect Document: I thought this was actually nicer, seems easier to inspect when you have a narrow pane.
I disagree. It's nicer if you know why it's missing, and know how to get it back (make the pane wider). But if let's say you're an editor and you're used to the horizontal scroll and you're working in the Studio. Behind the scenes someone is deploying `sanity@3.31.0` and the editor refreshes the page. The "Compare" button is gone, is the feature gone? Is it moved somewhere else? It's not intuitive that it's hidden due to the width. It's a bit more confusing if you're already comparing a revision as the dropdown no longer shows:
```
🕓 Review changes ✔︎
{} Inspect
```
Indicating that the "Review changes" view is active, it's just... gone 😅

[I'm against disabling buttons for similar reasons](https://uxdesign.cc/never-ever-disable-buttons-more-accessible-alternatives-5f415d79ce40), the TL;DR is that it's bad for usability to hide/disable controls.
A better alternative would be to show an alternate screen, for example something like this:
Frame 2


> Could potentially workaround and prevent the collapsed state unless inspecting comments or AI assist? Would that be preferable?

Workaround it how?

rdunk:
But the "Inspect Document" feature still exists right? When you mentioned that it had odd behaviour, I assumed that to be it switching from being contained in the document pane to overlaying both pane and preview (which is what I preferred). Maybe I misunderstood?

I totally agree with you on the Compare button being missing being less than ideal! I will follow up with Studio DX about that.

stipsan:
> But the "Inspect Document" feature still exists right? When you mentioned that it had odd behaviour, I assumed that to be it switching from being contained in the document pane to overlaying both pane and preview (which is what I preferred). Maybe I misunderstood?


Aaah, I misunderstood, I took it as "Review changes" lol. I like the extra space sure. But it's odd that it's the only modal that does that and it feels like a bug. I don't think I'd want all modals to behave that way, as it blocks me from being able to interact in the preview while the modal is open 🤔 


> I totally agree with you on the missing Compare button being less than ideal. I will follow up with Studio DX about that.


Gotcha :)

> And the workaround I had in mind was potentially that we could react to both the inspect state and size when calling `setLayoutCollapsed`. I'd have to check if this would work ofc.

Ah allright. It feels like something that should be handled in Structure Tool though? As it's a matter of how well it responds to sizing constraints and doesn't feel like a case where Presentation and Structure needs to handle the panes UI differently 🤔

rdunk:
Cool, all valid points.

The document inspector seems to behave in this way because scroll works a bit differently in the collapsed state. The grayed/overlayed section in these images corresponds to the scrollable area:

Screenshot 2024-03-06 at 15 21 13

Screenshot 2024-03-06 at 15 21 18

I'd have to do some investigating to work out change needed in the studio to allow us to fix this on the Presentation side.

If we can get the "Review changes" option to surface, would you be happy merging this and logging the Document inspector as an issue? I feel that would be preferable to the current overflowing behaviour.

stipsan (approved):
I'm happy with merging this now and fixing in a follow up. The new behaviour isn't perfect but it's still better than the current one 👍

PR #1084, opened 2024-03-08
feat(svelte-loader): support passing single object to `useQuery`

rdunk:
A proposal for a little tweak to `useQuery` in the Svelte loader. Adds the ability to optionally pass a single object instead of three params.

Previously:
```ts
export let data: PageData;
const { query, params, options } = data;
const query = useQuery(query, params, options);
$: ({ data: product } = $query);
```

Now:
```ts
export let data: PageData;
const query = useQuery(data);
$: ({ data: product } = $query);
```


stipsan (approved):
Sweet!

PR #1195, opened 2024-03-21
fix: improve vite compat with styled-components

stipsan:
Seeing a lot of smoke around issues like https://github.com/styled-components/styled-components/issues/4275 and it seems the best way to protect against it is to never do:
```tsx
import styled from 'styled-components'
```
and instead always use the named export:
```tsx
import {styled} from 'styled-components'
```

I've tried adding a guard in `@sanity/pkg-utils` that automatically protects against it, but got blocked by: https://github.com/rollup/rollup/issues/5265


rdunk:
Just JS things

rdunk (approved):

PR #1208, opened 2024-04-01
fix: use stega options in loadQuery

SimeonGriggs:
Currently when passing `staga` config into `options` it doesn't actually get used by `loadQuery`.

This has not been tested. I'm specifically unsure if the JSON.stringified version actually works...


packages/react-loader/src/createQueryStore/universal.ts
```diff
@@ -40,6 +40,12 @@ export const createQueryStore = (
     params: QueryParams = {},
     options: Parameters<QueryStore['loadQuery']>[2] = {},
   ): Promise<QueryResponseInitial<QueryResponseResult>> => {
+    let stega: StegaConfig = {enabled: false}
+    if (options.stega) {
+      stega = options.stega === true ? {enabled: true} : options.stega
+    } else if (unstable__serverClient.instance?.config().stega) {
+      stega = unstable__serverClient.instance?.config().stega
+    }
```
stipsan, line 48: The `stega` option should only use the boolean short hand, as options given to `unstable__cache.instance.fetch` has to be strictly JSON serializable, and `stega.studioUrl`, `stega.filter`, and `stega.logger`, aren't.
```suggestion
    const stega = options.stega ?? unstable__serverClient.instance?.config().stega.enabled ?? false
```

stipsan (changes requested):
For `stega` to apply properly it needs to be extracted here: https://github.com/sanity-io/visual-editing/blob/1b248a7704b4226a5e4c950e8b9831e75b5a98c9/packages/core-loader/src/index.ts#L121

And passed in here: https://github.com/sanity-io/visual-editing/blob/1b248a7704b4226a5e4c950e8b9831e75b5a98c9/packages/core-loader/src/index.ts#L122-L127

packages/react-loader/src/createQueryStore/universal.ts
SimeonGriggs: So then the type for `options` on 41 is incorrect? It suggests you can pass a `StegaConfig`

stipsan: Those `options` were typed for usage with `perspective: "previewDrafts"`, which doesn’t use `async-cache-dedupe` under the hood and thus isn’t limited to inputs that can be deserialized from strings and thus can use all the same stega options as `client.fetch`. 

I don’t know how to make TypeScript differentiate between the two but I’m open for suggestions

SimeonGriggs: A discriminated union?

dnlmzw:
@SimeonGriggs Are there any plans to merge this PR?

rdunk:
I've just merged this https://github.com/sanity-io/visual-editing/pull/1873 for covering the same situation in the Svelte loader, it also happily happens to cover the exact request @stipsan made in this PR to `@sanity/core-loader`.

@SimeonGriggs want me to pick this up (resolve conflicts) and get it merged, or would you like to make the changes and I'll review + merge?

SimeonGriggs:
> @SimeonGriggs want me to pick this up (resolve conflicts) and get it merged, or would you like to make the changes and I'll review + merge?

Argh I missed this!

I've just updated the original PR with Cody's recommendation, and merged in your updates. If you could review would be great, thanks!

packages/react-loader/src/createQueryStore/universal.ts
rdunk: Has anything changed here? The latest version seem to include the suggested removal, is this safe to do now?

SimeonGriggs: I gave up trying to fix the typings (this is why I abandoned the PR the first time).

Oh and yeah I screwed up and merged the wrong suggestion 🤦 however Cody's suggestion above still potentially creates a `StegaConfig` type

```ts
const stega = options.stega ?? unstable__serverClient.instance?.config().stega.enabled ?? false
```

Screenshot 2024-11-26 at 15 21 20

I'm wondering now if it's supposed to be:

```ts
const stega = options.stega ? unstable__serverClient.instance?.config().stega.enabled : false
```

Screenshot 2024-11-26 at 15 22 06

rdunk: Yep you're totally right about `StegaConfig` still being a possibility with the suggestion.

If we want to normalize to a boolean, can we just do this:

```suggestion
const stega =
  typeof options.stega === 'boolean'
    ? options.stega
    : (options.stega?.enabled ??
      unstable__serverClient.instance?.config().stega?.enabled ??
      false)
```

SimeonGriggs: It's filthy but I like it

rdunk: Please, just one more nested ternary nullish coalescence... I can stop whenever I want to.

@stipsan Is this correct behaviour to normalize to a bool? We good to go?

stipsan: Yes! Allowing other stega options on the request level opens up a world of pain 😮‍💨

rdunk (approved):
240 days later, LGTM!

rdunk (approved):
Let's try this again.

PR #1238, opened 2024-04-02
fix(visual-editing): prevent focusing both clicked and focused elements

rdunk:
I'm not sure how I didn't pick up on this one before, or if this is a result of the focus path changes I made in core, but it seems it's currently possible to focus two disparate overlay elements:

1. Click an element in the app preview.
2. The corresponding field gains focus in the document pane.
3. Focus a **different** field in the document pane.
4. There will be two active overlay elements (the clicked element does not lose focus).

Looks like this was a side effect of attempting to capture "reflected" events that occur when clicking on an element. Clicked elements were just left in place, regardless of whether or not the reflected event had a matching id/path.

This PR fixes that, so focusing on a different field should correctly drop focus of the previously clicked element.

### DIFF
--- a/packages/visual-editing/src/ui/elementsReducer.ts
+++ b/packages/visual-editing/src/ui/elementsReducer.ts
@@ -77,20 +77,27 @@ export const elementsReducer = (
         return { ...e, focused: false }
       })
     case 'presentation/focus': {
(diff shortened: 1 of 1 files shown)


stipsan (approved):
💖

PR #1382, opened 2024-04-22
fix(presentation): adjust toolbar item sizes and spacing

mariuslundgard:
Adjust the toolbar UI to align with UI changes to Structure with the introduction of Sanity UI v2.


rdunk (approved):
Nice. Feels much better aligned with the rest of the UI.

stipsan (approved):
Very nice indeed ✨

PR #1388, opened 2024-04-23
feat(presentation): add resolver APIs

rdunk:
This PR introduces two new APIs for resolving locations and documents. Both are optionally provided via the top level `resolve` plugin option.

#### `resolve.locations`

An improved Locations API which (soft) deprecates the `locate` option.

It accepts a `DocumentLocationResolver` as `locate` does, but also accepts a simplified object format which abstracts away observable logic, meaning users no longer need to wrangle with RxJS.

#### `resolve.mainDocuments`

Accepts an array of `DocumentResolver` objects, which contain a path pattern (using `path-to-regexp`) to match against the current previewUrl, and a method for resolving a document: a document type, groq filter, or resolver function.

### Todo before merging
- [x] Documentation


rdunk:
Docs are nearly there, so we can probably review this, hopefully it can land in the next release.

stipsan (approved):
Really well designed and implemented, no notes! I want to merge, release and test this in our templates 💖 ✨ 🥳  We should be able to make it for the release cycle of `sanity` next week 🤞

PR #1392, opened 2024-04-24
feat(presentation): combine viewport buttons into single toggle

rdunk:
I'm dropping this here for consideration but I've _slightly_ unconvinced myself because of the classic flip-flop button issue.

The PR combines the two viewport buttons into a single toggle. That means we only get to display one icon at a time.

When in desktop mode, do we show the desktop icon (current state) or the mobile icon (anticipated state on click)? Does it matter?

### DIFF
--- a/packages/presentation/src/i18n/resources.ts
+++ b/packages/presentation/src/i18n/resources.ts
@@ -74,14 +74,12 @@ export default {
   /** Text describing the current status of the preview frame */
   'preview-frame.status_timeout':
     'Unable to connect, check the browser console for more information.',
-  /** The `aria-label` for the button that switches to a full viewport */
-  'preview-frame.viewport-full-button.aria-label': 'Full viewport',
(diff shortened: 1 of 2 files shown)


stipsan (dismissed):
If we make it show the anticipated state (mobile) when the desktop viewport size is active then it works the same way as `sanity-plugin-iframe-pane`. I think it's nice for editors that migrate from the iframe pane and to presentation to have less things to "re-learn" ☺️

mariuslundgard (dismissed):

rdunk:
History related problems are upstream and should be fixed by [this PR](https://github.com/sanity-io/sanity/pull/6923). Keeping two icons for now as @mariuslundgard and I discussed.

This one is ready for final review by anyone now.

stipsan (approved):
🚢 🚢 🚢 💨 💨 💨

PR #1412, opened 2024-04-25
fix: panels should recover from broken local storage state

stipsan:
Sometimes the localStorage for panel sizes are corrupted, and it causes Presentation to render a WSOD since all panels effectively has zero width:
Screenshot 2024-04-25 at 21 17 27

I don't have a consistent way of reproducing it, or an explanation for how it stores `null` values to begin with. Since I had a local repro I was able to test the fix though and confirm it is able to restore default panel settings when the localStorage state has `null` instead of numbers.

### DIFF
--- a/packages/presentation/src/panels/usePanelsStorage.ts
+++ b/packages/presentation/src/panels/usePanelsStorage.ts
@@ -20,7 +20,7 @@ const getKeyForPanels = (panels: PanelElement[]) => {
 }
 
 export function usePanelsStorage(): {
-  get: (panels: PanelElement[]) => number[]
+  get: (panels: PanelElement[]) => number[] | undefined
   set: (panels: PanelElement[], widths: number[]) => void
   setDebounced: (panels: PanelElement[], widths: number[]) => void
 } {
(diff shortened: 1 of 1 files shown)


rdunk:
This seems to reset the panel widths on each app initialisation rather than keeping the localStorage values? Do you see that too?

Also the tool crashes if no `presentation/panels` value exists as `stored[key]` is undefined, the type casting I did in `getStoredItem` is inaccurate.

Do you have an example of some corrupted data I could test with?

stipsan:
Ooops, good catch @rdunk! I guess it was a bit too late night 😅 Pushed a fix.

You can use this to reproduce the corrupted state, [on this route](https://visual-editing-studio.sanity.build/page-builder-demo/presentation?preview=/):
```
localStorage.setItem('presentation/panels', JSON.stringify({'navigator:1,preview:3,content:5': [null, null, null], 'preview:3, content:5': [65.02219557352286, 34.97780442647713]}))
location.reload()
```
And you should now see the same thing. Again, I don't know how that corrupted state is written there to begin with. I just know that there have been multiple times were I've experienced the WSOD issue, then cleared my localStorage state and have it recover. I didn't dig further into it as I sometimes mess around with that state and cause problems to myself. This time though I knew for a fact that I haven't done any manual changes to localStorage, nor touched any code that read/writes to it so I suspected we have a real bug.
Another pattern is that it typically only happens when I swap from one of the other workspaces and to the Page Builder Demo one in the Visual Editing Studio 🤔

rdunk (approved):
New fix looks good to me! Can't seem to reproduce any problems.

PR #1414, opened 2024-04-26
fix: revert react peer dep to ^18.2.0

stipsan:
The bump of peer deps in #1409 of `react` to `^18.3.0` was a mistake. We don't actually require it, the only difference between 18.3 and 18.2 are deprecation warnings. Requiring 18.3 creates unnecessary churn later as `@sanity/presentation` is a dependency of `sanity` and it'll effectively require all users of `sanity` to also upgrade to `18.3` and all we'll gain from doing so is frustrating our users 😅

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -36,7 +36,10 @@
         "next",
         "vue-router",
         "nuxt",
-        "@remix-run/react"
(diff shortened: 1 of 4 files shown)


rdunk (approved):
Cool, makes sense!

PR #1418, opened 2024-04-26
fix: the intial `preview` param no longer "sticks" until page reload

stipsan:
When opening a route, like https://visual-editing-studio.sanity.build/next/app-router?preview=/404, the `preview=/404` state would "stick". You could navigate to other tools and workspaces, click on `Presentation` in the navbar, and it would keep loading the same initial `/404` route instead of falling back to the default route.
With this PR it now "unsticks", and as a side-effect if you click on the tool in the navbar it does a reset, which is useful in dev if the app you're previewing crashed.


rdunk (commented):
Assuming we could do a similar thing using `suspend-react`'s `clear` function, but I guess this approach means we don't have to wrangle with deps when calling that.

So looks good, much better behaviour, happy to merge!

rdunk (approved):

PR #1419, opened 2024-04-26
chore: fix page-builder-demo SSR hydration mismatch error

stipsan:
Fixes hydration mismatch errors that frequently happened when working locally.
Also uses `loadQuery` to speedup load perf.


rdunk:
I'm seeing the same error locally as the failed page builder demo instance related to `experimental_taintUniqueValue`. Is this a `18.3.0` thing?

rdunk (approved):
LGTM!

PR #1437, opened 2024-05-01
fix(visual-editing): maintain overlay element registration when toggling

rdunk:
Toggling "Edit" on and off currently causes the overlay controller instance to essentially be destroyed and recreated, meaning visibility of overlay elements is "lost" when in the off state.

This PR changes the behaviour so that the controller instance is maintained, along with the state which tracks elements, even when "Edit" mode is off. Instead just the elements themselves are activated and deactivated internally.

This has the side effect of preventing the documents on screen being reported each time "Edit" is toggled, which was in turn causing a loading state to be repeatedly displayed in the Document list.


stipsan (approved):
Thank you, this has been an annoying pet peeve to me 🙇

PR #1451, opened 2024-05-02
feat(presentation): support i18n and add default locale resources

rdunk:
First foray into using the i18n API so might require some testing! 😄

The string descriptions are also tedious, so feel free to suggest improvements.


stipsan (approved):
Wow, that was fast! Awesome 💖

PR #1463, opened 2024-05-05
feat: open last published document when published perspective

stipsan:
Uses the new `prefersLastPublished` parameter introduced in https://github.com/sanity-io/sanity/pull/6486


christianhg:
Okay, I'd love to test this so I understand it.

Could you give me some simple steps? I've already got the repo running locally.

stipsan:
> Okay, I'd love to test this so I understand it.
> 
> Could you give me some simple steps? I've already got the repo running locally.

Change the perspective to "Published":
image

As you click around you should see the published version of the document, instead of a draft.
Your experience should match this video:

https://github.com/sanity-io/visual-editing/assets/81981/1539d9e6-127f-469d-bb3e-2e24fa0e77b0

Compare that experience to what happens when you do the same interactions on [the `main` branch](https://visual-editing-studio.sanity.build) 🙌

packages/presentation/src/loader/BroadcastDisplayedDocument.tsx
```diff
@@ -1,20 +1,62 @@
...
   value: Partial<SanityDocument> | null | undefined
-}): null {
+}): JSX.Element | null {
   const setDisplayedDocument = useDisplayedDocumentBroadcaster()
+  const params = usePresentationParams(false)
 
-  useEffect(() => setDisplayedDocument?.(props.value), [props.value, setDisplayedDocument])
+  useEffect(() => {
+    if (params?.perspective !== 'published') {
+      const timeout = setTimeout(() => setDisplayedDocument?.(props.value), 10)
```
christianhg, line 24: Is there any sort of deeper meaning behind the 10ms timeout here and the 100ms timeout below?

stipsan: Well, technically we could remove the 10ms all together on this one. The timeout is only needed for the published document lookup as it takes a moment to load the `editState` :)

christianhg (approved):
I managed to test this locally and verify that it works 🎉

PR #1469, opened 2024-05-07
feat: add Astro example shoe store

christianhg:
A new `apps/astro` app has been added. I've tried to keep it as simepl as possible, and as close to the other shoe stores as I could. This means that the markup is copy/pasted with only very few changes (removal of `key` attributes, renaming `className` to `class`, etc). Some of the markup is rendered conditionally based on a loading flags, and even though these will always evaluate to `false` for now, I decided to keep them. When we add visual editing, we'll begin to need them. But for now, the pages are rendered using SSR with no need for loading states.


apps/common/src/env.ts
```diff
@@ -48,6 +48,12 @@ export const studioUrl = maybeGitBranchStudioUrl(
 export const apiVersion = '2023-10-11'
 
 export const workspaces = {
+  'astro': {
+    projectId: projectId,
+    dataset: datasets.development,
+    workspace: 'astro',
```
christianhg, line 54: @stipsan , we probably need to create this, right? 🤔

christianhg:
I've added https://vercel.com/sanity-io/visual-editing-astro. When this PR is merged, then we can adjust that project so it points to the correct root dir.

stipsan (approved):

stipsan (approved):
@christianhg sounds good! 👍

PR #1479, opened 2024-05-08
feat(studio): configure Astro presentation workspace URL

christianhg:
I've also configured the `SANITY_STUDIO_ASTRO_PREVIEW_URL` env var of the studio on Vercel to point to `https://visual-editing-astro.sanity.build/shoes`.

PR #1482, opened 2024-05-08
fix: regression causing a crash in embedded studios

stipsan:
In embedded Studios the `previewUrl` option typically won't have `origin` specified, leading to a crash:
image

Tested this fix on https://github.com/sanity-io/nextjs-blog-cms-sanity-v3/pull/505

I would write a test for it but I feel we should fast track a fix and ask Studio DX to do a patch of `sanity` and circle back later instead, since everyone on embedded studios are currently unable to use Presentation Tool 😅

### DIFF
--- a/packages/presentation/src/useMainDocument.ts
+++ b/packages/presentation/src/useMainDocument.ts
@@ -111,7 +111,7 @@ export function useMainDocument(props: {
       typeof previewUrl === 'string'
         ? previewUrl
         : typeof previewUrl === 'object'
-          ? previewUrl?.origin
+          ? previewUrl?.origin || location.origin
           : location.origin
 
     return new URL(relativeUrl, base)


packages/presentation/src/useMainDocument.ts
```diff
@@ -114,7 +114,7 @@ export function useMainDocument(props: {
           ? previewUrl?.origin
           : location.origin
 
-    return new URL(relativeUrl, base)
+    return new URL(relativeUrl, base || location.origin)
```
christianhg, line 117: I would probably handle the case in the expression above.

christianhg (approved):

christianhg:
@stipsan , the astro build fails. If we merge https://github.com/sanity-io/visual-editing/pull/1479 then you can rebase on `main` and get that build to go green.

christianhg (approved):

Issue #1486, opened 2024-05-08
Presentation breaks with Error: Maximum update depth exceeded. 


mdillgen:
**Describe the bug**
Latest update breaks presentation and throws an error immediately!

**To Reproduce**
https://github.com/sanity-io/visual-editing/issues/1486#issuecomment-2103758981   ⬇️

**Which versions of Sanity are you using?**

```
@sanity/cli (global)                   3.41.1 (up to date)
@sanity/document-internationalization   3.0.0 (up to date)
@sanity/image-url                       1.0.2 (up to date)
@sanity/preview-url-secret             1.6.12 (up to date)
@sanity/react-loader                   1.9.17 (up to date)
@sanity/types                          3.41.1 (up to date)
@sanity/ui                              2.1.6 (up to date)
@sanity/vision                         3.41.1 (up to date)
@sanity/visual-editing                 1.8.18 (up to date)
sanity                                 3.41.1 (up to date)
```

**What operating system are you using?**
macOS 14.4.1 

**Which versions of Node.js / npm are you running?**

Run `npm -v && node -v` in the terminal and copy-paste the result here.
```
npm 10.1.0
Node v20.9.0
```

ryanleichty:
I'm running into the same issue with the latest version.

![cleanshot_2024-05-07_at_21 30 40_2x](https://github.com/sanity-io/visual-editing/assets/17421429/6b5121e6-875a-4c20-9290-6bab4a98b973)

christianhg:
Thank you for reporting this. Can you elaborate a bit on your reproduction steps. In what repository did you reproduce it?

mdillgen:
> Thank you for reporting this. Can you elaborate a bit on your reproduction steps. In what repository did you reproduce it?

Hey @christianhg ,

Thank you for your reply, and sorry for the lack of detail in **the reproduction steps:**

1️⃣ To be fair, I set up a fresh repo, forked from the great template-nextjs-personal-website:
[https://github.com/mdillgen/template-nextjs-personal-website](https://github.com/mdillgen/template-nextjs-personal-website)

2️⃣ Updated dependencies
❌ Error occurs!

_Note that I also fixed the deprecated locate in the config, as you can see in the latest commit. BUT the error still happens._

ryanleichty:
[...]

PR #1496, opened 2024-05-13
fix(presentation): remove use of startTransition

christianhg:
When investigating #1486 inside
https://github.com/sanity-io/template-nextjs-personal-website I found that the `liveQueries` in `PresentationTool` easily run into an infinite loop causing the component to crash due to `Maximum update depth exceeded`. When inspecting the changes to the `liveQueries` that caused the re-renders I didn't find any differences expect a slight difference in `receivedAt`. Proceeding with the investigation, I couldn't find any odd places where `setLiveQueries` was called or where the `liveQueries` might be mutated. However, as soon as I remove the use of `startTransition`, when expiring the queries at a regular interval, the problem goes away. The component, `PresentationTool`, suddenly feels completely stable and nothing I do in the UI can provoke the infinite loop again. I'm not, however, sure how `startTransition` could cause this issue or what implications it might have that we remove it. All I know is that it fixes the issue at hand.

### DIFF
--- a/packages/presentation/src/PresentationTool.tsx
+++ b/packages/presentation/src/PresentationTool.tsx
@@ -15,7 +15,6 @@ import {BoundaryElementProvider, Flex} from '@sanity/ui'
 import {
   lazy,
   type ReactElement,
-  startTransition,
   Suspense,
   useCallback,
   useEffect,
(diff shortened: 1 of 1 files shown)


stipsan (approved):
Said template uses `next`, which uses Canary versions of `react`: https://react.dev/blog/2023/05/03/react-canaries#example-react-server-components
It's possible that the behaviour of `startTransition` has changed in the canary version of `react` that `next` is using.
For example the [`Caveats` section](https://react.dev/reference/react/startTransition#caveats) of the `startTransition` docs mentions this:
> If there are multiple ongoing Transitions, React currently batches them together. This is a limitation that will likely be removed in a future release.

Such a change would certainly explain what's happening, since the other state updates in `PresentationTool` leans heavily on the [Automatic Batching that React 18 introduced](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching).

That said the implications of removing the `startTransition` wrapper here is minimal. We used to also perform data fetching with Suspense, and marking the `liveQueries` state as a `transition` allowed these fetches to be interruptible and to run in the background. [Unfortunately Suspense for data fetching don't have a way of cancelling network calls](https://github.com/facebook/react/issues/21701) and so we refactored the `useQuerySubscription` hook to use `useEffect` as it allowed us to use `AbortController` to cancel requests that don't have a subscriber anymore.

It might be a good idea to look into other use cases where we have `startTransition` and see if we should remove those as well. 

Excellent work on investigating this! 🎉

PR #1504, opened 2024-05-14
refactor(ci): enable merge queues and automerge

stipsan:
Note that enabling automerge in renovatebot doesn't allow it to bypass required status checks or code review. It uses the same automerge feature as in the GitHub PR web ui :)
It also updates the default CI workflow to run on `merge_group`, which is required to enable merge queues that have the same required status checks as regular merge rules.


christianhg (approved):
So if I get this right, before this change, we would have to manually merge renovate PRs, but now, they will be queued and merged automatically as long as they pass our CI checks (including an approval)?

stipsan:
Yup!

PR #1521, opened 2024-05-15
feat(astro): add visual editing POC

christianhg:
It's not much, but I think it's enough to get the ball rolling on visual editing in Astro.

All queries are now passed through a `loadQuery` function similar to the one found in the Next.js server-only example. Here, we set the proper config to support visual editing in case that env var is set to `'true'`. Similarly, a `<VisualEditing />` component is added to the root layout in order to toggle visual editing if enabled. The `<VisualEditing />` component just triggeres a full page refresh whenever a refresh should happen.

Some questions that pop up are:

- What is the best way to toggle visual editing in Astro? Is it an env var like we use here, or is there a better way a la `draftMode` in Next.js?
- How do you best encapsulate the `<VisualEditing />` component to make sure its script is only loaded and run when visual editing is enabled? In the current implementation, `enableVisualEditing` is only called if the env var flag is set, but the `@sanity/visual-editing` module is still included, which might not be ideal. I've tried different techniques outlined in https://docs.astro.build/en/guides/client-side-scripts/, and the data-attribute-passed-to-custom-component is the technique I found to work the best. But maybe I missed something.
- Is there a better way to refresh the page on changes than using `window.location.reload()`?

---

Test it locally:

1. Linking the Vercel project and pull the env vars
2. `pnpm dev:astro`

I'll try to figure out how to get the deployment preview to work.

### DIFF
--- a/apps/astro/astro.config.mjs
+++ b/apps/astro/astro.config.mjs
@@ -15,10 +15,9 @@ export default defineConfig({
       dataset,
       useCdn: true,
       apiVersion,
-      studioUrl: ({_dataset}) =>
-        _dataset === workspaces['cross-dataset-references'].dataset
-          ? `${baseUrl}/${workspaces['cross-dataset-references'].workspace}`
-          : studioUrl,
+      stega: {
+        studioUrl,
+      },
     }),
     tailwind(),
   ],
--- a/apps/astro/src/components/visual-editing.astro
+++ b/apps/astro/src/components/visual-editing.astro
@@ -1,0 +1,50 @@
+---
+const visualEditingEnabled = import.meta.env.SANITY_VISUAL_EDITING_ENABLED
+---
+
+<sanity-astro-visual-editing enabled={visualEditingEnabled}></sanity-astro-visual-editing>
+
+<script>
+  import {enableVisualEditing} from '@sanity/visual-editing'
+
+  class SanityAstroVisualEditing extends HTMLElement {
+    static observedAttributes = ['enabled']
+    private unsubscribe: (() => void) | null = null
+
+    constructor() {
+      super()
+    }
+
+    disconnectedCallback() {
+      this.teardownVisualEditing()
+    }
+
+    attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null) {
+      if (name === 'enabled' && oldValue !== newValue) {
+        this.updateVisualEditing(newValue === 'true')
+      }
+    }
+
+    private updateVisualEditing(enabled: boolean) {
+      if (enabled) {
+        this.unsubscribe = enableVisualEditing({
+          refresh: () => {
+            return new Promise((resolve) => {
+              window.location.reload()
+              resolve()
+            })
+          },
+        })
+      } else {
+        this.teardownVisualEditing()
+      }
+    }
+
+    private teardownVisualEditing() {
+      this.unsubscribe?.()
+      this.unsubscribe = null
+    }
+  }
+
+  customElements.define('sanity-astro-visual-editing', SanityAstroVisualEditing)
+</script>
--- a/apps/astro/src/load-query.ts
+++ b/apps/astro/src/load-query.ts
@@ -1,0 +1,34 @@
+import {type QueryParams} from 'sanity'
+import {sanityClient} from 'sanity:client'
+
+const visualEditingEnabled = import.meta.env.SANITY_VISUAL_EDITING_ENABLED === 'true'
+const token = import.meta.env.SANITY_API_READ_TOKEN
+
+export async function loadQuery<QueryResponse>({
+  query,
+  params,
+}: {
+  query: string
+  params?: QueryParams
+}) {
+  if (visualEditingEnabled && !token) {
+    throw new Error('The `SANITY_API_READ_TOKEN` environment variable is required in Draft Mode.')
+  }
+
+  const perspective = visualEditingEnabled ? 'previewDrafts' : 'published'
+
+  const {result, resultSourceMap} = await sanityClient.fetch<QueryResponse>(query, params ?? {}, {
+    filterResponse: false,
+    perspective,
+    resultSourceMap: visualEditingEnabled ? 'withKeyArraySelector' : false,
+    stega: visualEditingEnabled,
+    ...(visualEditingEnabled ? {token} : {}),
+    useCdn: !visualEditingEnabled,
+  })
+
+  return {
+    data: result,
+    sourceMap: resultSourceMap,
+    perspective,
+  }
+}
--- a/apps/astro/package.json
+++ b/apps/astro/package.json
@@ -17,6 +17,7 @@
     "@sanity/astro": "^3.0.0",
     "@sanity/client": "^6.18.2",
     "@sanity/image-url": "^1.0.2",
+    "@sanity/visual-editing": "workspace:^",
     "apps-common": "workspace:^",
     "astro": "^4.8.4",
     "astro-portabletext": "^0.10.0",
--- a/apps/astro/src/env.d.ts
+++ b/apps/astro/src/env.d.ts
@@ -1,3 +1,12 @@
 /// <reference path="../.astro/types.d.ts" />
 /// <reference types="astro/client" />
 /// <reference types="@sanity/astro/module" />
+
+interface ImportMetaEnv {
+  readonly SANITY_VISUAL_EDITING_ENABLED: string
+  readonly SANITY_API_READ_TOKEN: string
+}
+
+interface ImportMeta {
+  readonly env: ImportMetaEnv
+}
--- a/apps/astro/src/layouts/layout.astro
+++ b/apps/astro/src/layouts/layout.astro
@@ -1,4 +1,6 @@
 ---
+import VisualEditing from '../components/visual-editing.astro'
+
 export type props = {
   title: string
 }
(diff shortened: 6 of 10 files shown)


apps/astro/astro.config.mjs
```diff
@@ -15,10 +15,9 @@ export default defineConfig({
       dataset,
       useCdn: true,
       apiVersion,
-      studioUrl: ({_dataset}) =>
-        _dataset === workspaces['cross-dataset-references'].dataset
-          ? `${baseUrl}/${workspaces['cross-dataset-references'].workspace}`
-          : studioUrl,
+      stega: {
+        studioUrl,
+      },
```
christianhg, line 20: I had to change this because `@sanity/astro` doesn't support a function here (I've shared more info about this in Slack). Also, `studioUrl` is now **inside** the `stega` config.

stipsan: Could you link to the slack thread? 🙌

stipsan: We don't have support for cross dataset references yet in Presentation anyway, so we can revisit this later once we land that :)

christianhg: > Could you link to the slack thread? 🙌

Here you go! https://sanity-io.slack.com/archives/C02CRE5GS2K/p1715715049501499

apps/astro/src/load-query.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  return sanityClient.fetch<QueryResponse>(query, params ?? {}, {
+    filterResponse: false,
+    perspective: isDraftMode ? 'previewDrafts' : 'published',
+    resultSourceMap: isDraftMode ? 'withKeyArraySelector' : false,
+    stega: isDraftMode,
+    ...(isDraftMode ? {token} : {}),
+    useCdn: !isDraftMode,
+  })
```
stipsan, line 19: I would change this to follow the same pattern as `@sanity/react-loader`'s and `@sanity/svelte-loader`'s implementations:

```suggestion

  const perspective = isDraftMode ? 'previewDrafts' : 'published'
  const {result, resultSourceMap} = await sanityClient.fetch<QueryResponse>(query, params ?? {}, {
    filterResponse: false,
    perspective,
    resultSourceMap: isDraftMode ? 'withKeyArraySelector' : false,
    stega: isDraftMode,
    ...(isDraftMode ? {token} : {}),
    useCdn: !isDraftMode,
  })
  return {data: result, sourceMap: resultSourceMap, perspective}
```

apps/astro/src/components/visual-editing.astro
stipsan, line 1: I would consider updating the custom element to tear down the overlays if the element is removed from the page, and if the prop itself is toggled (super handy for debugging as you can just toggle the attribute in the inspector). I also recommend using the term Preview Mode instead of Draft Mode as we reserve Draft Mode for when we specifically talk about Next.js.

```ts
---
const enabled = import.meta.env.SANITY_VISUAL_EDITING_ENABLED
---

<visual-editing enabled={enabled}></visual-editing>


<script>
  import { enableVisualEditing } from '@sanity/visual-editing';

  class VisualEditing extends HTMLElement {
    private unsubscribe: (() => void) | null = null;

    static get observedAttributes() {
      return ['enabled'];
    }

    constructor() {
      super();
    }

    connectedCallback() {
      this.updateVisualEditing();
    }

    disconnectedCallback() {
      this.teardownVisualEditing();
    }

    attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null) {
      if (name === 'enabled' && oldValue !== newValue) {
        this.updateVisualEditing();
      }
    }

    private updateVisualEditing() {
      this.teardownVisualEditing();

      if (this.enabled === 'true') {
        this.unsubscribe = enableVisualEditing({
          refresh: () => {
            return new Promise((resolve) => {
              window.location.reload();
              resolve();
            });
          },
        });
      }
    }

    private teardownVisualEditing() {
      if (this.unsubscribe) {
        this.unsubscribe();
        this.unsubscribe = null;
      }
    }
  }

  customElements.define('visual-editing', VisualEditing);
</script>
```

stipsan (commented):
>What is the best way to toggle visual editing in Astro? Is it an env var like we use here, or is there a better way a la draftMode in Next.js?


Since there is no first class primitive, similar to Next.js' Draft Mode, I don't think we have any other option but to toggle by env var unfortunately 🤷 

>How do you best encapsulate the <VisualEditing /> component to make sure its script is only loaded and run when visual editing is enabled? In the current implementation, enableVisualEditing is only called if the env var flag is set, but the @sanity/visual-editing module is still included, which might not be ideal. I've tried different techniques outlined in https://docs.astro.build/en/guides/client-side-scripts/, and the data-attribute-passed-to-custom-component is the technique I found to work the best. But maybe I missed something.

The `enableVisualEditing` function itself already implements lazy loading internally, so it should be fine to import `enableVisualEditing` statically.
The script is minimal: https://esm.sh/v135/@sanity/visual-editing@1.8.19/X-dHMvZW5hYmxlVmlzdWFsRWRpdGluZw/es2022/visual-editing.mjs
To reproduce the compilation above, use https://esm.sh/@sanity/visual-editing?exports=enableVisualEditing, it gives you an idea of what the initial bundle looks like for something that imports only `enableVisualEditing`, where the rest is loaded on demand.
While looking this up I noticed our lazy loading could be better so I'll make a quick follow up PR after this review 🙌 


>Is there a better way to refresh the page on changes than using window.location.reload()?

Not yet, not that I know. This could change dramatically if Astro chose to adopt React Server Components so we could revisit then.

apps/astro/src/load-query.ts
christianhg: Good idea!

apps/astro/src/components/visual-editing.astro
christianhg: Excellent suggestions. Will look into this.

christianhg:
@stipsan , is `feat` the correct conventional commit prefix, or is `chore` better?

apps/astro/src/components/visual-editing.astro
christianhg: > I also recommend using the term Preview Mode

What about just "Visual Editing" like the env var name?

stipsan: > What about just "Visual Editing" like the env var name?

It would read as this?
```html
<visual-editing enabled={visualEditingEnabled}></visual-editing>
```
That makes sense, and is much better actually since I just remembered we reserve calling things `Preview Mode` when we talk about the `presentationTool.previewUrl.previewMode` feature, and the related `@sanity/preview-url-secret` handshake 🤔 

Btw maybe we should call it `<sanity-visual-editing>`, it's been a while since I worked with web components but I think it's best practice to vendor prefix them :)

stipsan:
> @stipsan , is `feat` the correct conventional commit prefix, or is `chore` better?

If we want to be pedantic then `feat` is correct. The pragmatic answer is that the prefix you use only matters when it's used to generate release notes for `CHANGELOG.md` files, GitHub releases and to increment version numbers in `package.json` files. If the commit isn't touching any of those then it's ok to use `chore`, and it's appreciated if the prefix makes semantic sense since it makes it easier to know what the commit does at glance (is it a refactor? docs? a bugfix? adding functionality?) 😌

christianhg:
> > @stipsan , is `feat` the correct conventional commit prefix, or is `chore` better?
> 
> If we want to be pedantic then `feat` is correct. The pragmatic answer is that the prefix you use only matters when it's used to generate release notes for `CHANGELOG.md` files, GitHub releases and to increment version numbers in `package.json` files. If the commit isn't touching any of those then it's ok to use `chore`, and it's appreciated if the prefix makes semantic sense since it makes it easier to know what the commit does at glance (is it a refactor? docs? a bugfix? adding functionality?) 😌

Thanks! I also thought `feat` was correct here, but I'm new to conventional commits and I became a bit in doubt when I noticed my use of `feat` here https://github.com/sanity-io/visual-editing/pull/1469/commits/6429f25c6547463e03fdffc5cd946d157819dfd6 someone got changed into `chore`: https://github.com/sanity-io/visual-editing/commit/6fb2f005980de0fcf183e2754c7c9f36aa4a80d9

😅

stipsan:
> > > @stipsan , is `feat` the correct conventional commit prefix, or is `chore` better?
> 
> > 
> 
> > If we want to be pedantic then `feat` is correct. The pragmatic answer is that the prefix you use only matters when it's used to generate release notes for `CHANGELOG.md` files, GitHub releases and to increment version numbers in `package.json` files. If the commit isn't touching any of those then it's ok to use `chore`, and it's appreciated if the prefix makes semantic sense since it makes it easier to know what the commit does at glance (is it a refactor? docs? a bugfix? adding functionality?) 😌
> 
> 
> 
> Thanks! I also thought `feat` was correct here, but I'm new to conventional commits and I became a bit in doubt when I noticed my use of `feat` here https://github.com/sanity-io/visual-editing/pull/1469/commits/6429f25c6547463e03fdffc5cd946d157819dfd6 someone got changed into `chore`: https://github.com/sanity-io/visual-editing/commit/6fb2f005980de0fcf183e2754c7c9f36aa4a80d9
> 
> 
> 
> 😅 

Sure, good thing you don't have to worry about it if the commit isn't touching files within a workspace that is setup in release please manifest 😌 or lerna/semantic-release 😄

christianhg:
@stipsan , as a POC, I wonder if this PR is good enough to merge? Next up would be to find out how to package up a proper solution for the community.

stipsan (approved):
Yes, good enough to merge 👍

PR #1526, opened 2024-05-15
refactor: improve tsconfig settings

stipsan:
Based on the setup in the https://github.com/sanity-io/sanity monorepo.


christianhg (approved):
Sweet. I really don't like `baseUrl`.

PR #1528, opened 2024-05-15
fix: improve lazy loading of `enableVisualEditing`

stipsan:
Improves lazy loading by shifting more code into the lazy loaded function itself.

[Before](https://esm.sh/@sanity/visual-editing@1.8.20?exports=enableVisualEditing):
```js
/* esm.sh - esbuild bundle(@sanity/visual-editing@1.8.20) es2022 production */
import{jsx as i,Fragment as m}from"/stable/react@18.3.1/es2022/jsx-runtime.js";var u="sanity-visual-editing",e=null,t=null,r=null;function a(l={}){r&&clearTimeout(r);let n=new AbortController;return Promise.all([import("/v135/react-dom@18.3.1/es2022/client.js"),import("/v135/@sanity/visual-editing@1.8.20/X-dHMvZW5hYmxlVmlzdWFsRWRpdGluZw/es2022/dist/_chunks-es/VisualEditing.js")]).then(([o,{VisualEditing:s}])=>{if(!n.signal.aborted){if(e||(e=document.createElement("div"),e.id=u,document.body.appendChild(e)),!t){let{createRoot:d}="default"in o?o.default:o;t=d(e)}t.render(i(m,{children:i(s,{...l})}))}}),()=>{n.abort(),r=window.setTimeout(()=>{t&&(t.unmount(),t=null),e&&(document.body.removeChild(e),e=null)},1e3)}}export{a as enableVisualEditing};
```
This import causes a lot of code to load even though `enableVisualEditing` isn't called yet:
```
import{jsx as i,Fragment as m}from"/stable/react@18.3.1/es2022/jsx-runtime.js"
```

[After](https://esm.sh/@sanity/visual-editing@1.8.21-canary.0?exports=enableVisualEditing)
```js
function e(o={}){let r=new AbortController;return import("/v135/@sanity/visual-editing@1.8.21-canary.0/X-dHMvZW5hYmxlVmlzdWFsRWRpdGluZw/es2022/dist/_chunks-es/renderVisualEditing.js").then(({renderVisualEditing:n})=>{let{signal:t}=r;t.aborted||n(t,o)}),()=>{r.abort()}}export{e as enableVisualEditing};
```
Much better, the only import is the dynamic `import("/dist/_chunks-es/renderVisualEditing.js")`, and the surrounding code is minimal 🎉

### DIFF
--- a/packages/visual-editing/src/ui/enableVisualEditing.tsx
+++ b/packages/visual-editing/src/ui/enableVisualEditing.tsx
@@ -1,11 +1,4 @@
-import type {Root} from 'react-dom/client'
-
-import {OVERLAY_ID} from '../constants'
 import type {DisableVisualEditing, VisualEditingOptions} from '../types'
-
-let node: HTMLElement | null = null
-let root: Root | null = null
-let cleanup: number | null = null
 
 /**
  * Enables Visual Editing overlay in a page with sourcemap encoding.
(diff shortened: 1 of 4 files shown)


packages/visual-editing/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sanity/visual-editing",
-  "version": "1.8.20",
+  "version": "1.8.21-canary.0",
```
christianhg, line 3: Can you explain this change to me?

packages/visual-editing/src/ui/renderVisualEditing.tsx
```diff
@@ -0,0 +1,52 @@
+/**
+ * The purpose of this file is to contain the logic for rendering the <VisualEditing />
+ * component in a way that is easy to lazy load for the `enableVisualEditing` function.
+ */
```
christianhg, line 4: Great idea!

packages/visual-editing/package.json
stipsan: I just ran a canary release locally so I could compare the output on esm.sh as used in the PR description :)

stipsan: I didn't mean to commit that though hang on

packages/visual-editing/src/ui/enableVisualEditing.tsx
```diff
@@ -1,58 +1,32 @@
...
+    /**
+     * Due to lazy loading it's possible for the following to happen, and is a consequence of dynamic ESM imports not being cancellable natively:
+     * 1. Userland calls `const disableVisualEditing = enableVisualEditing()` and the dynamic ESM import is started.
+     * 2. The dynamic import uses the network, and it takes a while to load.
+     * 3. The user navigates to a different page in the app that doesn't need Visual Editing, for example a login page.
+     * 4. Since the app is no longer in a state where Visual Editing is needed, the user calls `disableVisualEditing()`.
+     * 5. The dynamic import eventually resolves and this function is called.
+     * When this happens we want to skip calling `renderVisualEditing` since we know it's no longer needed.
+     */
+    if (signal.aborted) return
```
christianhg, line 23: Perfection ✨

packages/visual-editing/package.json
stipsan: Fixed

christianhg (approved):

PR #1530, opened 2024-05-15
fix: group all internal `sanity` imports in one location

stipsan:
By grouping them in a single location we can link to it from here: https://github.com/sanity-io/sanity/blob/7ae3afd32d119b6e85508e3d1810489b67f6133e/.github/CODEOWNERS#L72-L97 and it's easier to maintain the list.

Also fixes the deprecation notices in TS related to these: https://github.com/sanity-io/sanity/blob/7ae3afd32d119b6e85508e3d1810489b67f6133e/packages/sanity/src/structure/index.ts#L1-L22 so that Studio DX can delete these the next time they update their `@sanity/presentation` version.


packages/presentation/src/internals.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * The purpose of this file is to gather all the internals we use from `sanity` in a single location.
+ * It makes maintenance easier, and allows us to quickly report to the Studio DX teams which internals we use, and
+ * what changes they can make without it becoming breaking.
+ */
```
christianhg, line 5: For some reason I'm on the fence about this change. To me it adds two downsides:

1. An added layer of indirection when reading code. ("Where does this come from? Internals? What's that...")
2. Manual work needed in order to adhere to this pattern (unless I misunderstood something)

stipsan: I understand that. At the same time yesterday when trying to find out "what are all the internal Sanity APIs we use?" I spent hours digging through everything and it won't be the last time we need to answer that question, as Studio DX needs to be in the loop with what files they can stop exporting or refactor 😅

stipsan: The manual work to keep this file up to date is much smaller than the repeated work of analyzing this information on demand. Additionally we're not intending on staying on internals forever, the plan is to eventually make the APIs we use stable and have the "@public" tsdoc tags, be officially safe to use and supported. So we intend to remove the need to have this file you could say

christianhg: It sounds really smart that we can link to the file directly from `CODEOWNERS` in the other repo. But what does it mean in practice? Does it mean that our team can now patch, review and merge these internals ourselves?

christianhg: Okay, I think I get it. So the imports that you've left behind, like `SanityClient` are the `@public` ones?

christianhg (approved):

packages/presentation/src/internals.ts
stipsan: > It sounds really smart that we can link to the file directly from `CODEOWNERS` in the other repo. But what does it mean in practice? Does it mean that our team can now patch, review and merge these internals ourselves?

It means that if anyone changes files that have internals that can't change without possibly breaking Presentation then our team is automatically requested to review a PR. 
Studio DX can still chose to merge the PR without our approval, for example some files contain multiple internal exports and we might only be using one or two of them, and Studio DX might know ahead of time that they're not touching any exports we're actually using.
Even if it's merged before we manage to review it we'll still be notified of the PR and given the opportunity to see the change and if we suspect it will be breaking we have a good chance to flag it and notify the Studio DX team and prevent it from going out in a new release before we've tested a canary of the change and verified it isn't breaking 🙌

If our team opens a PR that changes those internals we'll still need Studio DX's approval before merge ☺️

stipsan: > Okay, I think I get it. So the imports that you've left behind, like `SanityClient` are the `@public` ones?

Exactly 😌

PR #1545, opened 2024-05-17
feat(visual-editing): add Astro VisualEditing component

christianhg:


christianhg:
I'll add this to `@sanity/astro` instead.

PR #1559, opened 2024-05-23
chore(app-astro): use VisualEditing component from @sanity/astro

christianhg:


christianhg:
The build fails with this error on Vercel, which is quite strange because it works locally when I test and the prop isn't required.

Screenshot 2024-05-23 at 12 12 07

christianhg:
> The build fails with this error on Vercel, which is quite strange because it works locally when I test and the prop isn't required.
> 
> Screenshot 2024-05-23 at 12 12 07

I was able to reproduce this locally by recreating what is happening on Vercel:

1. `cd apps/astro`
2. `pnpm install`
3. `turbo run build`

Now, the `@sanity/visual-editing` dependency can't be resolved from inside `@sanity/astro` which caused the prop type error. I've mitigated this by manually specifying the `@sanity/visual-editing` dependency in the app. I suspect this error won't happen in user land, and it has more to do with this monorepo setup. What do you think, @stipsan ?

stipsan:
> > The build fails with this error on Vercel, which is quite strange because it works locally when I test and the prop isn't required.
> > Screenshot 2024-05-23 at 12 12 07
> 
> I was able to reproduce this locally by recreating what is happening on Vercel:
> 
> 1. `cd apps/astro`
> 2. `pnpm install`
> 3. `turbo run build`
> 
> Now, the `@sanity/visual-editing` dependency can't be resolved from inside `@sanity/astro` which caused the prop type error. I've mitigated this by manually specifying the `@sanity/visual-editing` dependency in the app. I suspect this error won't happen in user land, and it has more to do with this monorepo setup. What do you think, @stipsan ?

Yep this is similar to what we do in the other apps that use wrappers. For example both `dev/next-no-cache` and `dev/next-server-only` uses `import {VisualEditing} from 'next-sanity'` and have to use this in their package json as well:
```json
"@sanity/visual-editing": "workspace:*"
```
And just like with `next-sanity` userland shouldn't need to do this :)

stipsan (approved):

christianhg:
> > > The build fails with this error on Vercel, which is quite strange because it works locally when I test and the prop isn't required.
> > > Screenshot 2024-05-23 at 12 12 07
> > 
> > 
> > I was able to reproduce this locally by recreating what is happening on Vercel:
> > 
> > 1. `cd apps/astro`
> > 2. `pnpm install`
> > 3. `turbo run build`
> > 
> > Now, the `@sanity/visual-editing` dependency can't be resolved from inside `@sanity/astro` which caused the prop type error. I've mitigated this by manually specifying the `@sanity/visual-editing` dependency in the app. I suspect this error won't happen in user land, and it has more to do with this monorepo setup. What do you think, @stipsan ?
> 
> Yep this is similar to what we do in the other apps that use wrappers. For example both `dev/next-no-cache` and `dev/next-server-only` uses `import {VisualEditing} from 'next-sanity'` and have to use this in their package json as well:
> 
> ```json
> "@sanity/visual-editing": "workspace:*"
> ```
> 
> And just like with `next-sanity` userland shouldn't need to do this :)

Fantastic! Thanks for reassuring me!

PR #1570, opened 2024-05-28
fix!: move `react` and `react-dom` to peer deps

stipsan:
Hopefully this fixes the build issues in Astro 🙌


christianhg (approved):
This is probably the way to go. Bring your own version of React to get Visual Editing to work.

PR #1580, opened 2024-05-29
fix(types): add react export to typeVersions

christianhg:
Otherwise the types are not properly exported.


stipsan (approved):
Good catch!

christianhg:
> Good catch!

I found out the hard way 😆

PR #1596, opened 2024-05-29
fix: add firefox support to "Copy URL" action

mariusGundersen:
This should make the button also work in Firefox where ClipboardItem isn't implemented. The change keeps the ClipBoardItem working as before for iOS but relies on a promise that resolves to the text that should be copied in Firefox. This way both browsers work.


christianhg (approved):
Thank you for this! It looks good!

PR #1630, opened 2024-06-06
feat: add new @sanity/insert-menu package

christianhg:
In order to expose this menu to both `@sanity/visual-editing` and `sanity`.

This repository might not be the best permanent resident for this package, but that's OK for now.

---

I used `pnpx @sanity/pkg-utils@latest init my-package` to bootstrap this which was rather nice. After bootstrapping, I adjusted it a little bit based on how some of the other packages are set up in this repo. But mostly it's just `pkg-utils`.

**The `InsertMenu` code is more-or-less copy-pasted from my PR here: https://github.com/sanity-io/sanity/pull/6853. I've in turn repurposed that PR to use the `InsertMenu` from this package. Therefore, the best way to test this PR is by spinning up `pnpm dev:page-building-studio` on https://github.com/sanity-io/sanity/pull/6853**

I had to copy/paste a few utility functions in order to get `getSchemaTypeIcon` working.

When this PR is merged, I expect that I can adjust https://github.com/sanity-io/sanity/pull/6853 to import the menu from this new package and then use it.


packages/insert-menu/src/InsertMenu.tsx
```diff
@@ -0,0 +1,302 @@
...
+export type InsertMenuProps = InsertMenuOptions & {
+  schemaTypes: Array<SchemaType>
+  onSelect: (schemaType: SchemaType) => void
+  labels: {
+    'insert-menu.filter.all-items': string
+    'insert-menu.search.no-results': string
+    'insert-menu.search.placeholder': string
+    'insert-menu.toggle-grid-view.tooltip': string
+    'insert-menu.toggle-list-view.tooltip': string
+  }
```
christianhg, line 76: @stipsan Is this somewhat OK for labels? On the call site these are populated using translatable strings: https://github.com/sanity-io/sanity/pull/6853/files#diff-0f3eb80086c6f4701bf8c9bf9f95cee729761baf5a8a1dcb75e375b63d3ed25dR141-R147

packages/insert-menu/src/InsertMenu.tsx
```diff
@@ -0,0 +1,302 @@
...
+                  position: 'absolute',
+                  left: '50%',
+                  top: '50%',
+                  transform: 'translateX(-50%) translateY(-50%)',
+                },
+              })
+            : null}
+          {failedToLoad ? null : (
+            <img
+              src={`/static/preview-${props.schemaType.name}.png`}
```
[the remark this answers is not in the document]
christianhg: I made a fixup that changes the `views` config a bit to introduce a required `previewUrl` callback on the grid view: e820b45caa029c00ae024b547c30326173fa33fb

packages/insert-menu/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist",
+    "src"
+  ],
+  "scripts": {
+    "build": "pkg build --strict --clean --check",
+    "lint": "eslint .",
+    "ts:check": "tsc --build"
```
stipsan, line 37:
```suggestion
    "ts:check": "tsc --noEmit"
```
Oops we don't use incremental tsc builds anymore, we use turborepo instead 🙌

packages/insert-menu/src/InsertMenu.tsx
```diff
@@ -0,0 +1,297 @@
...
+  type MenuItemProps,
+  Stack,
+  Tab,
+  TabList,
+  Text,
+  TextInput,
+  Tooltip,
+} from '@sanity/ui'
+import {type ChangeEvent, createElement, useReducer, useState} from 'react'
+import {isValidElementType} from 'react-is'
```
stipsan, line 19: Since `react-is` is used here we need to add it to peer deps and dev deps the same way we specify `react` and `react-dom` 🤔

We should probably go over our other packages as I suspect a lot of our npm packages are doing this wrong 😅

packages/insert-menu/src/InsertMenu.tsx
stipsan: Yeah, nice and pragmatic 💖

stipsan (commented):
Looks great! Look at the react-is comment and lmk what you think 🙌

packages/insert-menu/src/InsertMenu.tsx
christianhg: Ah, of course. I should have caught that because I was actually looking in the `dist` folder to verify that we didn't bundle React and friends, but somehow I missed this one. `dist` is much cleaner now.

packages/insert-menu/package.json
christianhg: Fixed: 588a3971f53d1e12073b5f9a8617d7d819bc3d83

packages/insert-menu/src/InsertMenu.tsx
christianhg: Fixed: 588a3971f53d1e12073b5f9a8617d7d819bc3d83

stipsan (approved):

PR #1646, opened 2024-06-10
fix(presentation): dedupe search param derived state

rdunk:
This PR removes perspective and viewport state from presentationReducer's `PresentationState`. This state is already stored as URL search parameters. Removing this duplicate state means we don't have to use `useEffect` to keep search parameters in sync, we just use the URL as the source of truth.

This resolves issues we were seeing with browser history where moving back would actually cause new history to be pushed to the stack. Note that this issue still persists when viewing published documents as the studio internally pushes a `rev` parameter when `prefersLatestPublished` is present. AFAICT this is a temporary solution so this is acceptable for now.

Using `<StateLink>`s will also now be more predictable, and any state passed will more accurately reflect the actual state of the tool after navigation has finished.

Would appreciate if someone (@stipsan?) could verify that removing perspective and viewport from `PresentationState` will not have any unwanted consequences that I haven't forseen!

#### Note

To display the Main Document, we now use `SanityDefaultPreview` over the standard `Preview` component. This is because these changes mean the link parameters for the main document preview now update in the same way as those in the document list (previously this link did not contain the viewport size search parameter). Using `<Preview>` is no longer possible as switching viewport size or perspective would cause a re-render and a flash of skeleton content whilst that component resolves the required document information.


stipsan (approved):
All good! I don't think removing `perspective` and `viewport` from the reducer will have any regressions :) 

Also wanted to add that perhaps we could consider taking ownership of `prefersLastPublished`: https://github.com/sanity-io/sanity/blob/576dd3faae3b7d09079f95810a569e32870c69c0/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx#L151-L164
Then we can remove that code from `sanity`, and fully own it e2e. It feels like we can solve the `when in perspective: published selecting the "latest" version in the revision dropdown is reset to a revision` issue if we do 🤔

rdunk:
Sounds like a good idea re: `prefersLastPublished`, would be good to hear more about it as I was out when that decision was made. Either way this PR should play nicely with it for now, and fix the jumping to the incorrect revision issue we were seeing.

PR #1652, opened 2024-06-11
fix(presentation): support references in location resolver selections

rdunk:
This PR adds missing support for following references in `select` strings when defining Document Locations Resolvers to better align with the Preview API.

For example, defining locations for a `Book` document, given `author` is a reference type:
```ts
defineLocations({
  select: {
    title: 'title',
    author: 'author.name',
    slug: 'slug.current',
  },
  resolve: (doc) => ({
    locations: [
      {
        title: [doc?.title, doc?.author].filter(i=>i).join(' by ') || 'Untitled',
        href: `/books/${doc?.slug}`,
      },
      {
        title: 'Books',
        href: '/books',
      },
    ],
  }),
})
```


packages/presentation/src/util/props.ts
rdunk, line 1: Note that this is lifted [from here](https://github.com/sanity-io/sanity/blob/18aee900adeebc4e0df2fd71e751c93f323bc4c8/packages/sanity/src/core/preview/utils/props.ts#L4).

stipsan (commented):
Sweet! 💖

stipsan (approved):

PR #1657, opened 2024-06-13
fix(insert-menu): ui polish

mariuslundgard:


christianhg:
It looks good. We have an issue where it sometimes end up here, but it doesn't look like that is introduced by this PR.

Screenshot 2024-06-13 at 14 14 06

christianhg (approved):
Thank you! 👏

christianhg:
> It looks good. We have an issue where it sometimes end up here, but it doesn't look like that is introduced by this PR.
> 
> Screenshot 2024-06-13 at 14 14 06

Never mind, I've fixed it now in https://github.com/sanity-io/sanity/pull/6853

PR #1664, opened 2024-06-14
fix(insert-menu): improve InsertMenuOptions.filter docs

christianhg:


rdunk (commented):
Excessively nitpicking, feel free to ignore as I think it's fine as is.

Would you consider "Filtering is automatically enabled if [..]"? As technically the `filter` option itself is not changed.

Would it be helpful to explicitly state what happens on `filter: true`, or is that implied/obvious?

christianhg:
> Excessively nitpicking, feel free to ignore as I think it's fine as is.
> 
> Would you consider "Filtering is automatically enabled if [..]"? As technically the `filter` option itself is not changed.
> 
> Would it be helpful to explicitly state what happens on `filter: true`, or is that implied/obvious?

I appreciate any help with the wording that I can get.

We are also considering a different API, using enums instead of a three-way boolean, to make it even more clear what the different options are and what they do:

```
filter?: 'auto' | 'on' | 'off' // `'auto'` is default
// or
filter?: 'auto' | 'enabled' | 'disabled' // `'auto'` is default
```

What do you think?

rdunk:
```
filter?: 'auto' | 'on' | 'off' // `'auto'` is default
```

Yeah great, an enum is much nicer IMO. I like the one above because `on`/`off` to me more obviously implies "always on" or "always off", where as `auto` and `enabled` seem quasi-synonymous in this context.

christianhg:
> ```
> filter?: 'auto' | 'on' | 'off' // `'auto'` is default
> ```
> 
> Yeah great, an enum is much nicer IMO. I like the one above because `on`/`off` to me more obviously implies "always on" or "always off", where as `auto` and `enabled` seem quasi-synonymous in this context.

Thanks a lot! I'll pivot this PR to use `auto`/`on`/`off`.

PR #1679, opened 2024-06-18
fix(insert-menu): adjust API and auto-capitalise group.name/schemaType.name if used as title fallback

christianhg:


stipsan (approved):
Go go go 🏎️💨

rexxars (approved):
LGTM

PR #1683, opened 2024-06-19
fix(insert-menu): set `placement` property of tooltip

mariuslundgard:
Fixes a typo in the `placement` property of the tooltip.


christianhg (approved):
Thank you 😬

PR #1691, opened 2024-07-02
fix(presentation): prevent displaying main document on explicit document navigation

rdunk:
### Overview

This one has been reported by a few different people.

Currently the main document resolver is far too keen in automatically navigating users to the main document for a given path:

A main document is resolved (or not) for the current preview path after each, and on the initial, navigation. This means that if a user follows a link which contains a path which resolves a main document, but a document id that does not match it, the main document will override that specified.

This PR fixes this by comparing the previous two router states (i.e. the previous navigation event) before forcing a main document navigation. If the previous navigation did not trigger a document change, or in other words the path changed but the document did not, then we can safely show the main document.

### Test

🔴 [Navigating on main](https://visual-editing-studio.sanity.build/page-builder-demo/presentation/product/462efcc6-3c8b-47c6-8474-5544e1a4acde?preview=/) will show a flash of the document whose ID corresponds to that in the URL, and then show the main document.

🟢 [Navigating on this branch](https://visual-editing-studio-git-crx-795.sanity.build/page-builder-demo/presentation/product/462efcc6-3c8b-47c6-8474-5544e1a4acde?preview=%2F) will correctly display the document whose ID corresponds to that in the URL.

### DIFF
--- a/packages/presentation/src/useMainDocument.ts
+++ b/packages/presentation/src/useMainDocument.ts
@@ -2,7 +2,7 @@ import type {ResponseQueryOptions} from '@sanity/client'
 import {match, type Path} from 'path-to-regexp'
 import {useCallback, useEffect, useMemo, useRef, useState} from 'react'
 import {useClient} from 'sanity'
-import {useRouter} from 'sanity/router'
+import {type RouterState, useRouter} from 'sanity/router'
 
 import {API_VERSION} from './constants'
 import type {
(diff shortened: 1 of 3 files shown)


packages/presentation/src/useMainDocument.ts
```diff
@@ -157,11 +157,20 @@ export function useMainDocument(props: {
...
-                })
                 mainDocumentIdRef.current = doc?._id
+
+                // We only want to force a navigation to the main document if
+                // the path changed but the document ID did not. An explicit
+                // document navigation should take precedence over displaying
+                // the main document. We determine if an explicit document
+                // navigation has occured by comparing the IDs of the last two
+                // resultant navigation states.
+                if (navigationHistory.at(-1)?.['id'] === navigationHistory.at(-2)?.['id']) {
```
snorrees, line 168: Nit: cant you just `.id` here?

snorrees (approved):
LGTM!

packages/presentation/src/useMainDocument.ts
rdunk: I wish! But Presentation's tsconfig inherits `noPropertyAccessFromIndexSignature` from `@sanity/pkg-utils` presets, so we'd get a `Property 'id' comes from an index signature, so it must be accessed with ['id'].` error if we tried to do that AFAIK.

PR #1692, opened 2024-07-02
fix(presentation): preserve tool state across intent link navigations

rdunk:
### Summary

This pull request addresses the handling of router state when navigating via intent links in the Presentation tool.

Currently, the Presentation tool does not maintain search parameter-specific state when handling intent links. This can lead to unintended side effects, such as resetting the preview URL and changing the viewport size when intent links are used to navigate within the tool.

### Proposed Solution
The solution involves updating the `getIntentState` method to preserve the search parameters related to the Presentation tool's state (as defined in the [PresentationSearchParams](https://github.com/sanity-io/visual-editing/blob/main/packages/presentation/src/types.ts#L173-L177) interface), rather than requiring them to be explicitly defined when constructing an intent link.

### Testing and Validation
In our test environment, we have an experimental navigator component that uses intent links to create new documents in-situ. You'll notice a narrow/mobile viewport size will not be maintained when using these links because the relevant parameter is not explicitly set. On this branch, that state will be maintained.

Feedback is appreciated to confirm if this approach aligns with the expected behaviour of intent links. Is preserving these parameters the correct approach? Should the handling of intent links be tool/parameter specific, or is there a general expectation for how they should behave?

### DIFF
--- a/packages/presentation/src/getIntentState.ts
+++ b/packages/presentation/src/getIntentState.ts
@@ -1,8 +1,14 @@
 import {uuid} from '@sanity/uuid'
-import {type SearchParam} from 'sanity/router'
+import {type RouterState, type SearchParam} from 'sanity/router'
 
 import {encodeJsonParams, getPublishedId} from './internals'
-import type {PresentationStateParams} from './types'
+import type {PresentationSearchParams, PresentationStateParams} from './types'
+
+const preservedSearchParamKeys: Array<keyof PresentationSearchParams> = [
+  'preview',
+  'perspective',
+  'viewport',
+]
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


packages/presentation/src/getIntentState.ts
```diff
@@ -1,44 +1,60 @@
...
+  routerState: RouterState | undefined,
   payload: unknown,
 ):
   | (PresentationStateParams & {_searchParams: SearchParam[]})
   | {intent: string; params: Record<string, string>; payload: unknown} {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const {id, mode, path, presentation, type, ...searchParams} = params
 
+  const preservedSearchParams = (routerState?._searchParams || [])
+    // @todo Casting https://github.com/microsoft/TypeScript/issues/14520
```
stipsan, line 28: We can probably remove the casting after we upgrade to TS 5.5 😄

stipsan (approved):
I think it's justifiable.
I feel the current routing API in the Studio leaves something to be desired, let's follow up with the team over the summer 👍

packages/presentation/src/getIntentState.ts
rdunk: Great! 🥳

PR #1693, opened 2024-07-02
fix(presentation): add iframe pointer events to prevent interference with resizing drag handler

rdunk:
### Overview

We prevent users interacting with iframe content whilst the iframe is loading or whilst a connection is being established to  the Presentation tool.

Currently, this is done using `pointer-events: none;` on the iframe itself. However this means panel resizing does not play nicely at all (see Linear ticket for video), as I assume it prevents `mousemove` events as a user's cursor hovers over the iframe.

This PR attempts similar behaviour by replacing `pointer-events: none;` with a transparent absolutely positioned div, preventing interaction with the iframe but allowing the `mousemove` events we need to handle resizing.

Also some minor refactoring to tidy up the iframe component.

### Test

Drag the divider between iframe and document pane in Presentation tool on main and then this branch. Your dragging experience should be markedly improved.

### DIFF
--- a/packages/presentation/src/panels/PanelResizer.tsx
+++ b/packages/presentation/src/panels/PanelResizer.tsx
@@ -107,11 +107,35 @@ export const PanelResizer: FunctionComponent<{
   useEffect(() => {
     if (!isDragging || disabled) return
 
+    // Set styles to prevent text selection and force an ew-resize cursor whilst
+    // dragging. Return a reset callback so we can revert to any values that
+    // might have been present before dragging started.
+    function setDocumentStyles() {
+      const bodyStyle = document.body.style
+      const documentStyle = document.documentElement.style
+
+      const {cursor} = documentStyle
+      const {userSelect} = bodyStyle
+
+      documentStyle.cursor = 'ew-resize'
+      bodyStyle.userSelect = 'none'
+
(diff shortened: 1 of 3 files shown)


robinpyon (dismissed):
LGTM – tested both before and after and wasn't able to replicate the previous issue, much nicer!

Below are a few minor issues / nits I did discover on testing – they're not related to this PR, but thought worth mentioning (just in-case they're not already captured!)

- The resize cursor icon doesn't stay active when dragging 'wide' (when your mouse is no longer positioned over the divider). This could possibly be addressed by setting the `cursor` on `document.documentElement` directly (and decoupling that from the pane divider itself)
- It's possible to get the resize handle in a locked state by opening the context / right mouse button menu (real edge case though, and this also happens on the studio)
- When resizing, it's still possible to _select_ text in either the presentation or document panels. Similarly to the first point, this could possibly be addressed by conditionally setting `userSelect: none` on `document.body`, but only when it's currently dragging

rdunk:
Thanks so much for the thorough review and for offering suggestions, and in particular for detailing possible solves for them!

I think they are all great suggestions, so I've added two more commits to include them.

robinpyon (approved):

PR #1699, opened 2024-07-29
fix(presentation): display correct revision when switching perspective

rdunk:
This PR adds a small renderless component that handles switching to the correct document revision when the perspective changes. I believe this removes the need for `prefersLatestRevision`, and allows us to simplify the components and hooks that relied on it.


stipsan (dismissed):
Seems to work well, as far as I can see in my own testing 👍  💖

stipsan (approved):

PR #1700, opened 2024-07-29
fix(visual-editing): import svelte types from dist over src

rdunk:
Related to https://github.com/sanity-io/sanity-template-sveltekit-clean/issues/40.

Currently we import types from the sibling `/src` directory, which results in errors as `@repo/` imports cannot be resolved.


stipsan (approved):
❤

PR #1705, opened 2024-08-05
fix(visual-editing): improve overlay mutation handling

rdunk:
This adds a method for updating an existing element detected by the overlay controller, rather than having to de-and-re-register it to track updates.

It also adds support for handling text mutations, which means changes in stega encoded strings will be picked up (they weren't before!)

It seems to make things much more predictable: no more randomly missing overlays!


stipsan (approved):
Awesome fix 💖  Let's ship 🚀 🚀 🚀

PR #1720, opened 2024-08-07
chore(deps): upgrade to turbo v2, pnpm v9 & improve dev scripts

stipsan:
To improve dev scripts the new `turbo watch` command is used, it re-runs the `dev` commands in `packages/*` that currently runs `pkg build --strict`.

Once `@sanity/pkg-utils` handles TS types with its `watch` command we can move to a simplified setup with `turbo run dev` again instead of `turbo watch dev`, that handles running `build` commands during warm up. The reason we manually run `build` before `turbo watch` is due to `turbo` running `watch` tasks all in parallel, which means you easily have `sanity dev` happening before packages it uses, like `@sanity/presentation`, is built.

While at it I've bumped `@sanity/ui`, `sanity` packages, and enabled testing our `@sanity/styled-components` temporary fork ([related studio PR](https://github.com/sanity-io/sanity/pull/7330)).


.github/workflows/browserslist.yml
stipsan, line 1: This action doesn't really work well with `pnpm`

.github/workflows/ci.yml
```diff
@@ -23,12 +22,11 @@ jobs:
       TURBO_TEAM: ${{ vars.TURBO_TEAM }}
     steps:
       - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v4
```
stipsan, line 25: v4 automatically supports `packageManager` in `package.json`, so the corepack command is no longer necessary

.github/workflows/preview.yml
```diff
@@ -23,7 +22,7 @@ jobs:
         with:
           app-id: ${{ secrets.ECOSPARK_APP_ID }}
           private-key: ${{ secrets.ECOSPARK_APP_PRIVATE_KEY }}
-      - uses: connor-baer/action-sync-branch@main
+      - uses: connor-baer/action-sync-branch@v1.1.0
```
stipsan, line 25: Don't know why renovate missed pinning this one, it's pinned now though

.github/workflows/release-please-workaround.yml
stipsan, line 1: This workaround is no longer needed, as Google fixed this back in May https://github.com/googleapis/release-please/pull/2281 🥳

apps/next-no-cache/next.config.mjs
```diff
@@ -1,9 +1,19 @@
...
 }
 
+const sanityExports = {}
+for (const key of Object.keys(sanityPkg.exports)) {
+  if (key === '.') continue
+  const subexport = path.join('sanity', key)
+  sanityExports[subexport] = requireResolve(subexport)
+}
+sanityExports['sanity'] = requireResolve('sanity')
+
```
stipsan, line 16: We'll ship a first class utility for this in `next-sanity` later. This feature is based on the new aliasing that `sanity dev` does since v3.53.0

apps/next-no-cache/sanity.config.ts
```diff
@@ -5,9 +5,9 @@
 
 import {visionTool} from '@sanity/vision'
 import {defineConfig} from 'sanity'
+import {presentationTool} from 'sanity/presentation'
```
stipsan, line 8: Thanks to the `link-workspace-packages = deep` in the root `.npmrc` we can use e2e tests that import from `sanity/presentation` and still have it use the workspace version instead of npm 🥳

turbo.json
```diff
@@ -1,32 +1,27 @@
...
     },
     "lint": {
-      "outputMode": "errors-only"
+      "outputLogs": "errors-only"
     },
     "preview": {
       "dependsOn": ["^build"]
     },
     "dev": {
-      "dependsOn": ["^build"],
```
stipsan, line 24: This will be added back once `@sanity/pkg-utils` can run a watcher that also handles `d.ts` generation

release-please-config.json
```diff
@@ -1,10 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
-  "bump-minor-pre-major": true,
-  "always-link-local": true,
```
stipsan, line 4: `true` is the default value

apps/studio/sanity.cli.ts
```diff
@@ -1,27 +1,21 @@
...
-        'apps-common': require.resolve('../common/src'),
-        '@sanity/presentation': require.resolve('../../packages/presentation/src'),
-        'sanity/_internal': require.resolve('sanity/_internal'),
-        'sanity/_singletons': require.resolve('sanity/_singletons'),
-        'sanity/cli': require.resolve('sanity/cli'),
-        'sanity/desk': require.resolve('sanity/desk'),
-        'sanity/router': require.resolve('sanity/router'),
-        'sanity/structure': require.resolve('sanity/structure'),
-        'sanity': require.resolve('sanity'),
-        'styled-components': require.resolve('styled-components'),
```
stipsan, line 23: The Studio itself is now automatically adding these aliases, so we no longer have to maintain them 💖

apps/studio/sanity.config.ts
```diff
@@ -44,30 +44,12 @@ const sharedSettings = definePlugin({
   },
 })
 
-// If we're on a preview deployment we'll want the iframe URLs to point to the same preview deployment
-function maybeGitBranchUrl(url: string) {
```
stipsan, line 48: This stuff was always very brittle.
Moving forward deployments will always point to the production URLs.
If a PR has changes that require the Visual Editing Studio to open deployments of Nuxt/Next/Etc from the same branch, that is done by adding environment variables for that branch on the env settings: https://vercel.com/sanity-sandbox/visual-editing-studio/settings/environment-variables

package.json
```diff
@@ -1,46 +1,39 @@
...
   "scripts": {
     "build": "turbo run build --filter='./packages/*'",
     "build:astro": "turbo run build --filter=./apps/astro",
-    "dev": "turbo run dev --filter=./apps/* --filter=!./apps/next-server-only --filter=!./apps/next-no-cache",
-    "dev:astro": "turbo run dev --filter=./apps/astro --filter=./apps/studio",
-    "dev:next": "turbo run dev --filter=./apps/next --filter=./apps/studio",
-    "dev:nuxt": "turbo run dev --filter=./apps/nuxt --filter=./apps/studio",
-    "dev:page-builder-demo": "turbo run dev --filter=./apps/page-builder-demo --filter=./apps/studio",
-    "dev:svelte": "turbo run dev --filter=./apps/svelte --filter=./apps/studio",
+    "dev": "turbo run build --filter='./packages/*' && turbo watch dev --filter='./packages/*' --filter=./apps/* --filter=!next-no-cache --filter=!next-server-only --filter=!next-with-i18n",
```
stipsan, line 7: Chaining `turbo run build` with `turbo watch dev` commands are only necessary until `@sanity/pkg-utils` can handle `.d.ts` generation in its `watch` command. Once that is supported we can go back to just `turbo run dev` 🙌

packages/core-loader/turbo.json
stipsan, line 1: Once `@sanity/pkg-utils` can handle type generation in its `watch` command we can delete these `turbo.json` files that only serve to tell turbo that during `turbo watch` these `dev` commands should rerun when the workspace changes.

packages/insert-menu/turbo.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "dependsOn": ["^build"]
```
stipsan, line 6: Since the root `turbo.json` declares that this package should be built, we have to override it here to avoid a recursing dependency that turbo doesn't allow.
This is necessary to begin with due to the root level `.npmrc` setting `link-workspace-packages = deep` which causes the `@sanity/insert-menu` that `sanity` installs to point to the workspace.
Since `sanity` itself isn't a workspace package turbo isn't able to infer that anything that depend on `sanity` should also make sure that `@sanity/insert-menu` is built, we have to manually specify it __somewhere__. Declaring it on the root level ensures that "it just works", and prevents this implementation detail of the monorepo from leaking into lots of affected `turbo.json`'s in the apps folder.

packages/overlays/package.json
```diff
@@ -102,7 +101,7 @@
     "root": true
   },
   "dependencies": {
-    "@sanity/visual-editing": "2.1.8"
+    "@sanity/visual-editing": "workspace:^"
```
stipsan, line 104: Using `workspace:^` here is intentional, we want the published version to be `^2.1.8` instead of an exact version, so we can stop publishing new versions of `@sanity/overlays` and remove it from the monorepo.

apps/next-no-cache/next.config.mjs
rdunk: Cool! Also TIL about Import Assertions

apps/studio/sanity.config.ts
rdunk: Will that be a manual process per branch?

I guess this will be a requirement for almost every PR unless I'm mistaken?

package.json
rdunk: That's great. For now the slightly slower startup time is definitely worth the HMR improvements.

apps/studio/sanity.config.ts
stipsan: Yes, but not for long, I'll make a github action that automatically sets them 🙌

rdunk (approved):
No obvious issues that I could find in the code changes. Tested locally and it seems work great, big improvement. I left one query about the per branch env vars but not a blocker, can discuss that later on.

Thanks!

PR #1751, opened 2024-08-12
chore: remove `@sanity/overlays` automation

stipsan:
Previous release automation meant that `@sanity/overlays` were published with an exact dependency range for `@sanity/visual-editing`. Which meant that for users of `@sanity/overlays` to receive bugfixes we had to continue publishing new versions of `@sanity/overlays` everytime we changed `@sanity/visual-editing`.
We have since upgraded our automation and we recently published a new version of `@sanity/overlays` that has the dependency specified as `"@sanity/visual-editing": "^2.1.9"`. This means that patches and minors of `@sanity/visual-editing` will be pulled in by projects on `@sanity/overlays`, and we no longer need to have automation for both releasing `@sanity/overlays`, as well as automatically marking new versions as deprecated on npm.


rdunk (approved):
Awesome!

PR #1754, opened 2024-08-15
chore: fix storybook build command

stipsan:
We have to do a bit of a workaround in the project settings atm:

image

This PR lets us set it back to just `turbo run build-storybook` 🥳


rdunk (approved):
Ah right, because this just depends on the package's own build command, LGTM.

PR #1758, opened 2024-08-21
chore: add core drag and drop events

georgedoescode:


packages/visual-editing/src/util/findSanityNodes.ts
```diff
@@ -176,3 +176,18 @@ export function findSanityNodes(
   }
   return elements
 }
+
+export function getSanityNodeArrayPath(sanityPath: string): string {
```
georgedoescode, line 180: this is quite a rough implementation currently. will test nested array items, etc, when working on the UI.

packages/visual-editing/src/util/dragAndDrop.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    handler({
+      type: 'overlay/updateDragCursorPosition',
+      x: mousePos.x,
+      y: mousePos.y,
+    })
+
+    const insertPosition = calcInsertPosition(mousePos, rects, flow)
+
+    if (JSON.stringify(insertPosition) !== JSON.stringify(dragState.insertPosition)) {
```
georgedoescode, line 292: this could be updated later to avoid using `JSON.stringify`

stipsan: As long as we're not sensitive to object key ordering might change, or array values, I think `JSON.stringify` is the most efficient way to compare them anyway?

stipsan (dismissed):
Great work! 💖 LFGO

packages/visual-editing/src/util/dragAndDrop.ts
georgedoescode: I naively assumed comparing the numeric values would be more efficient, but a quick benchmark suggests `JSON.stringify` is faster! 🎉 Also, key ordering shouldn't be an issue in this case so I'll leave this be 🚀

georgedoescode:
> Great work! 💖 LFGO

Thank you! 🚀 ❤️

packages/visual-editing/src/controller.ts
```diff
@@ -1,20 +1,27 @@
...
 import {getRect} from './util/getRect'
 
 const isElementNode = (target: EventTarget | null): target is ElementNode => {
   return target instanceof HTMLElement || target instanceof SVGElement
 }
 
+const dragState: DragState = {
+  status: 'idle',
+  insertPosition: null,
+}
```
rdunk, line 23: Feel like I'm missing something here but I'll ask the question anyway: What is the `dragState` useful for in this scope? Is the shared state between  `handleOverlayDrag` doing something I haven't picked up on?

Also, what is the `status` value used for? I can see that it's mutated both in this file and within the utility function, but to what end?

georgedoescode: Good question! Currently, it tracks `insertPosition` so that we only update/emit events when a new position in calculated. 

`status` will be set to `dragging` when we start the UI work on `mousedown` so that we can disable certain overlay behaviour during a drag sequence :)

rdunk: Would it make sense to add something like `dragStatus` to the `OverlayState` interface so we have a single place for storing overlay related state, and then track the previous `insertPosition` within the `handleOverlayDrag` function (that is if that doesn't need to be shared state?)

georgedoescode: That makes total sense. I have gone ahead and remove `dragState` here and added a simple `prevInsertPosition` to the `handleOverlayDrag` function. We can add the UI-specific events next if that works, and integrate with the `overlayState` reducer :)

rdunk (approved):
LGTM!

PR #1760, opened 2024-08-23
feat(comlink): init package

rdunk:
Initial version of the comlink package. Not consumed by anything yet so happy to iterate, just wanted to get something committed. 

I'm happy with the state machine logic generally, less confident about the `controller` logic (and some internal types in that file) which handles creating `channel` instances when sources are added for example. Should this be a state machine too?

Feel free to interrogate, nitpick, question the public API (refer to the playground app for use) and suggest name changes, etc!


packages/comlink/src/common.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const eventToMessage =
+  <T>(type: T) =>
+  (event: MessageEvent<ProtocolMessage>): {type: T; message: MessageEvent<ProtocolMessage>} => ({
+    type,
+    message: event,
+  })
+
+export const messageEvents$ = defer(() =>
+  fromEvent<MessageEvent<ProtocolMessage>>(window, 'message'),
```
stipsan, line 58: rxjs + xstate = ❤

packages/comlink/src/common.ts
```diff
@@ -0,0 +1,73 @@
...
+)
+
+export const listenActor = fromEventObservable(({input}: {input: ListenInput}) => {
+  return messageEvents$.pipe(
+    filter(listenFilter(input)),
+    map(eventToMessage(input.responseType)),
+    input.count
+      ? pipe(
+          bufferCount(input.count),
```
stipsan, line 67: I love this!!

packages/comlink/src/common.ts
stipsan: Maybe we should setup a share replay here with deferred unsubscribe to handle cases where the <VisualEditing> component is unmounted and mounted again? This happens for example when moving between App Router and Pages Router in Next.js 🤔

packages/comlink/src/request.ts
```diff
@@ -0,0 +1,216 @@
...
+      return output
+    },
+  })
+}
+
+// export const delayedRequestMachine = requestMachine.provide({
+//   delays: {
+//     initialTimeout: 500,
+//   },
+// })
```
stipsan, line 216: Clever 😏

packages/comlink/src/request.ts
```diff
@@ -0,0 +1,216 @@
...
+            filter(
+              (event) =>
+                event.data.type === MSG_RESPONSE &&
+                event.data.responseTo === input.requestId &&
+                !!event.source &&
+                input.sources.has(event.source),
+            ),
+            take(input.sources.size),
+          ),
+      ),
```
stipsan, line 81: So clean 🥲

packages/comlink/src/node.ts
```diff
@@ -0,0 +1,430 @@
...
+    })
+    return unsubscribe
+  }
+
+  const post = (data: WithoutResponse<S>) => {
+    actor.send({type: 'post', data})
+  }
+
+  const fetch = (data: WithoutResponse<S>) => {
+    const resolvable = Promise.withResolvers<S['response']>()
```
stipsan, line 406: I'm unsure if our pkg-utils stack is updated to automatically polyfill this yet, I'll make a note to check this before we start shipping comlink to production 😌

packages/comlink/src/node.ts
stipsan, line 1: This is just amazing, stellar work @rdunk!!! I've never been able to get such an easy to follow Birds Eye view of how our channels package actually works 🥲

packages/comlink/src/channel.ts
stipsan, line 1: Superb 💖

stipsan (approved):
I'm merging this 💖 we can bikeshed and iterate all we want moving forward, I see no reason to hold off 😄
I don't think the controller needs to be a state machine, unless we need machines it spawns to inter communicate or something else that is easier to orchestrate in a state machine 😌

I'd love for the playground to be converted to a storybook, similar to how we set things up in the visual-editing package. Storybooks can let you open iframes between stories, and the storybook format itself is portable and reusable. A story can work as a playground, it can be reused as internal docs (as stories can demonstrate different scenarios), it can be reused as vitest tests, reused as playwright tests, and has a lot less boilerplate than remix in a sub folder 😌
Your remix playground is super easy to convert though so you made the right choice 👍

PR #1762, opened 2024-08-27
fix(presentation): prevent openInStructure from linking to the presentation tool itself

marcusforsberg:
Hey Sanity team! 👋

On some of our projects, the Presentation tool is the only tool used to work with content; i.e. there's no Structure tool at all.
In this scenario, the "Open in Structure" action will end up linking back to the Presentation tool itself, so nothing happens when you click on it.

Example screenshot below. Our Presentation tool is titled "Live Edit" and the action says "Open in Live Edit", but that tool is already open so the button effectively does nothing.

Screenshot 2024-08-27 at 15 38 52

This PR solves that by excluding the Presentation tool when trying to find matches in `findStructureTool`. With this change, there's no action button shown at all, since there's no relevant tool to link to:

Screenshot 2024-08-27 at 15 42 48


stipsan (approved):
Thanks for the fix! 💖

PR #1764, opened 2024-08-30
chore: improve vscode monorepo onboarding

stipsan:
The `settings.json` file ensures folks don't have to deal with making sure TypeScript loads correctly.

And the `extensions.json` will show a prompt in VS Code based editors:
image

I have a **_lot_** of extensions I like, but I wanted to make sure the recommendations are only the most useful ones, especially for the tooling we use in this monorepo:::

image

They are the kind of extensions that folks often don't know to look for, but are huge timesavers.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,5 @@
 {
   "recommendations": [
-    "Nixon.env-cmd-file-syntax",
     "sanity-io.vscode-sanity",
     "joshbolduc.story-explorer",
     "bradlc.vscode-tailwindcss",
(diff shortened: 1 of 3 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
+  "recommendations": [
+    "Nixon.env-cmd-file-syntax",
```
rdunk, line 3: Massive nitpick but this one seems non-essential?

stipsan: Agreed, dropping

rdunk (approved):
Great!

PR #1789, opened 2024-09-10
chore(drag-and-drop-opt-out)

georgedoescode:
This PR updates the Drag and Drop API to be opt-out. 

It also simplifies the functionality, enabling drag-and-drop only on elements with an array path.


stipsan (approved):
All good 👍

PR #1809, opened 2024-09-10
fix: remove buggy `document.referrer` behaviour

stipsan:
[More info](https://linear.app/sanity/issue/CRX-929/presentation-has-buggy-documentreferrer-behaviour)

### DIFF
--- a/apps/studio/sanity.cli.ts
+++ b/apps/studio/sanity.cli.ts
@@ -20,6 +20,7 @@ export default defineCliConfig({
           'apps-common': require.resolve('../common/src'),
           '@sanity/presentation': require.resolve('../../packages/presentation/src'),
         },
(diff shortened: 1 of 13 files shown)


apps/studio/sanity.cli.ts
```diff
@@ -20,6 +20,7 @@ export default defineCliConfig({
           'apps-common': require.resolve('../common/src'),
           '@sanity/presentation': require.resolve('../../packages/presentation/src'),
         },
+        dedupe: [...(config.resolve?.dedupe || []), 'sanity'],
```
stipsan, line 23: This is only necessary because we're overriding `@sanity/presentation` above

rdunk (approved):
Awesome, LGTM!

PR #1811, opened 2024-09-10
chore: add canary release workflow

stipsan:
Automation for publishing canaries on the `canary` branch. 
Can be run using the "Run workflow" button here https://github.com/sanity-io/visual-editing/actions/workflows/release-canary.yml (the button won't show up until after this PR is merged to main), once this workflow is synced to the `canary` branch.

The workflow itself will always use the `canary` branch, to avoid mistakes. You can also run `pnpm release:canary` locally, if you have publishing rights in our npm org.

I've tested the workflow on this branch, here's the commit it pushed: https://github.com/sanity-io/visual-editing/commit/a3b1834d741d44f024cc379f53a42c833809fd8e

The packages it published has provenance, and resolved dependencies correctly: https://www.npmjs.com/package/@sanity/react-loader/v/1.10.6-canary.3 (you can see it depends on `@sanity/core-loader@1.6.22-canary.3`).
I verified it by running the install action on the "Are we react 19 yet?" repo: https://github.com/sanity-io/are-we-react-19-yet/pull/322/files
If there were any problems with how the package was published, then this PR would fail to be created.

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -1,0 +1,49 @@
+name: Release Canary
+
+on: workflow_dispatch
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read # for checkout
+
+jobs:
+  release-canary:
+    permissions:
+        contents: read # for checkout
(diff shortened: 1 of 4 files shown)


rdunk (approved):
Awesome, thank you!

PR #1817, opened 2024-09-12
chore(prevent-dnd-in-non-studio-context)

georgedoescode:


rdunk:
Could we pass the same common `inFrame` const we define in `VisualEditing.tsx` to the controller via `useController`?

```diff
- const controller = useController(rootElement, overlayEventHandler, !!inFrame)
+ const controller = useController(rootElement, overlayEventHandler, inFrame)
```

And alter the signature here (and I guess in `createOverlayController` too):

```diff
export function useController(
  element: HTMLElement | null,
  handler: OverlayEventHandler,
- preventDefault: boolean,
+ inFrame: boolean,
)
```

stipsan (approved):

PR #1857, opened 2024-09-24
Drag and Drop Minimap v1

georgedoescode:
This PR: 

- Implements minimap scaling 
- Implements a small prompt to show when minimaps are available 
- Tidies up existing drag and drop code, creating a new `geometry` utils file that we can use for other areas 
- Updates drag sequence logic so that minimaps can persist until the shift key is unpressed
- Simplifies inverse transform logic using native browser APIs
- Updates `renderVisualEditing` to place our div outside of `<body>` - this greatly simplifies handling scaling overlays while minimaps are enabled.

I went back and forth on emitting events for the minimap scaling and applying the transforms, etc in `Overlays`. It feels simpler to keep it in the handler for now, but I'm open to feedback! 

Next Steps: 

In the interest of moving forward, and being able to demo this tomorrow, I have split a few small tasks out: 

- Enable developers to opt-out of minimaps 
- Hide overlays during minimap scaling animation sequences (listeners already attached) 
- Allow developers to define a custom scale target for minimaps (?) 
- Disable click events on overlays while minimaps are open


stipsan (approved):
Excellent 💖

PR #1868, opened 2024-09-25
fix(visual-editing): Don't show "Open in Studio" offscreen when at top of screen

DAlperin:
For things like navbars or logos that might appear on the top of the screen, the overlay could be cut off. This moves the "Open in Studio" below the element in that case.

## Before:
Screenshot 2024-09-25 at 3 10 21 PM

## After:
Screenshot 2024-09-25 at 3 11 35 PM


DAlperin:
@rdunk I'm not sure who to ping for a review, but you seem like a good person to try :)

PR #1871, opened 2024-09-26
chore(drag-and-drop): drag handles, drag centroids, bugfixes

georgedoescode:


packages/visual-editing/src/controller.ts
```diff
@@ -312,6 +298,7 @@ export function createOverlayController({
       id,
       rect: getRect(element),
       sanity,
+      dragDisabled: element.getAttribute('data-sanity-disable-drag') === 'false',
```
rdunk, line 301: As I understand it `getAttribute()` will return `'true'` (as a string) if the data attribute value is empty, and `null` if the attribute is missing?

So this would evaluate to true (i.e. drag is disabled) when rendering the following?:
`data-sanity-disable-drag="false"`

georgedoescode: Ah, good catch. We need to check for `true` rather than `false`! 🕵️

packages/visual-editing/src/util/dragAndDrop.ts
```diff
@@ -412,38 +424,52 @@ export function handleOverlayDrag(
...
       // cleanup keyup after drag sequence is complete
       if (!mousedown) {
-        window.removeEventListener('keyup', handleKeyup)
+        clearInterval(rectsInterval)
+        removeListeners()
       }
     }
   }
 
+  const handleBlur = () => {
```
georgedoescode, line 453: Hardening to prevent studio focus changes from getting folks "stuck" in minimap views

packages/visual-editing/src/util/dragAndDrop.ts
```diff
@@ -307,7 +315,13 @@ export function handleOverlayDrag(
 
   if (!minimapScaleApplied) prescaleHeight = document.documentElement.scrollHeight
 
+  const rectsInterval = setInterval(() => {
```
georgedoescode, line 318: Updating rect state at an interval helps keep geometry fresh during a drag sequence

packages/visual-editing/src/util/findSanityNodes.ts
```diff
@@ -201,3 +207,34 @@ export function sanityNodesExistInSameArray(
 
   return getSanityNodeArrayPath(sanityNode1.path) === getSanityNodeArrayPath(sanityNode2.path)
 }
+
+export function resolveDragAndDropGroup(
```
georgedoescode, line 211: We may need this function elsewhere later 👀

packages/visual-editing/src/ui/ElementOverlay.tsx
```diff
@@ -160,8 +164,15 @@ const ElementOverlayInner: FunctionComponent<ElementOverlayProps> = (props) => {
 
       {title && (
         <Tab gap={1} paddingBottom={1}>
-          <Labels gap={2} padding={2}>
-            <Text size={1}>{Icon}</Text>
+          <Labels gap={3} padding={3}>
```
rdunk, line 167: Were these sizing adjustments discussed with anyone else? The spacing here now differs from the `Open in Studio` link. Happy to consider but we should probably align.

georgedoescode: Good shout! These UI updates are quite rough. I am planning on meeting with Marius to discuss a few ideas for taking it further/polishing. I'll make sure the 2 contexts align 👀

rdunk: Sounds good! Shall we maybe revert the spacing for now, get this PR merged and then handle those on a separate branch?

georgedoescode: Sounds good! Done :)

PR #1873, opened 2024-10-01
feat: add svelte loader `LiveMode`, improve stega handling

rdunk:
#### `<LiveMode>`

Adds a new `<LiveMode>` component export to `@sanity/svelte-loader` which offers improved UX for enabling live mode.

This follows the same pattern as adding the `<VisualEditing>` component from `@sanity/visual-editing/svelte` to enable visual editing, rather than having to call a function in `onMount`.

In order to ship Svelte components, the loader package is now built using `svelte-package` instead of `@sanity/pkg-utils`

#### Stega Handling

Currently, stega is always returned by `loadQuery` if it is enabled at the client config level. Now `loadQuery` will check if preview mode is disabled, and accordingly disable stega.

This required a small change to `@sanity/core-loader`, in order to pass stega configuration to the unstable cache fetcher.


packages/svelte-loader/src/LiveMode.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
```
georgedoescode, line 1: Could `LiveMode` be a little confusing moving forward as we are using `SanityLive` for the Live Content API stuff?

rdunk: I think this is a good point, my thoughts are that:
- Yes it could definitely be confusing!
- We already export a `useLiveMode` function in this loader and in the React loader, so the idea of "live" is already established.
- This component is really just a wrapper that calls `useLiveMode` and saves a bit of boilerplate. Is the idea of introducing a _component_ with this name the confusing part?
- I don't think we're 100% confirmed on what the new live APIs will look like, so I'm tempted to say we don't worry too much about that for now.
- I think we'll likely do a major update of the Svelte loader soonish to support Svelte 5, so we can rethink the APIs at that point.

georgedoescode: This makes sense! I agree that since we already have the concept of `liveMode` established, this should be all good. Let's roll with it like you say and see how things feel once the API naming is more established, etc :)

georgedoescode (approved):
Have checked over the best I can with my limited loaders knowledge, but this looks goods to me. I have also pulled the branch and tested:

- overlays in both tools when in Presentation ✅
- overlays in both standalone apps when preview mode is enabled ✅
- no overlays and no stega in the markup in both standalone apps when preview mode is disabled ✅

PR #1884, opened 2024-10-04
feat(visual-editing): add `useDocuments` hook

rdunk:
Exports a new consumable hook that could potentially be used in custom overlay components and for overriding drag and drop functionality.


packages/visual-editing/src/ui/optimistic-state/useDocuments.ts
```diff
@@ -0,0 +1,79 @@
...
+      snapshot: SanityDocument<T>
+    }
+  }
+
+  const mutateDocument = useCallback<DocumentMutate>(
+    (documentId, mutations, options) => {
+      const {commit = true} = options || {}
+
+      if (isEmptyActor(actor)) {
+        const inFrame = window.self !== window.top || window.opener
```
stipsan, line 53: Not a blocker, but should we add a helper in visual-editing-helpers that does this? 
`const inFrame = isInFrame()`

Or maybe we should call it `isPresentation` or `isPresentable` instead of `inFrame`?

stipsan (approved):
Nice! ✨

PR #1886, opened 2024-10-07
chore(drag-and-drop-ui-updates)

georgedoescode:


stipsan (approved):
The changes makes sense to me, let's go 🚀🚀🚀

PR #1910, opened 2024-10-14
DnD Perf/UI Updates

georgedoescode:
This PR: 

- Refactors `DragInsertMarker` to use inline styles vs styled components. SC was killing framerate in Safari during minimap sequences 
- Adds an internal `activeDragSequence` variable to the overlay controller. This is currently used to prevent `click` events from firing shortly after drag sequences end — this occasionally caused the studio side to focus at an awkward time, resulting in frame-rate drops and unwanted `scrollIntoView` triggers.
- Adds `onSequenceStart` and `onSequenceEnd` callbacks to our `handleDragAndDrop` function, these are currently used to support the above, but could later be used to focus an item's parent array on drag start/end. 
- Always renders skeleton rects. We used to only render them for drag previews of a certain size, but since we remove the drag handle it is no longer necessary. 
- Uses a simple `lerp` to pick the best fitting radius value from the sanity UI design system for overlay drag previews
- Fixes preview rect stacking for elements with background images — we now always render image rects first, and text rects second.


packages/visual-editing/src/ui/OverlayDragPreview.tsx
```diff
@@ -58,22 +58,30 @@ const Root = styled.div<{
...
-  const showSkeleton =
-    skeleton.w * scaleFactor >= minSkeletonWidth && skeleton.h * scaleFactor >= minSkeletonHeight
-
   const offsetX = skeleton.offsetX * scaleFactor
   const offsetY = skeleton.offsetY * scaleFactor
 
   const prefersDark = usePrefersDark()
   const theme = useTheme_v2()
 
+  const radius = theme.radius[~~map(skeleton.w, 0, 1920, 1, theme.radius.length - 2)]
```
georgedoescode, line 80: it would be great to be able to do this in CSS alone... it would also be cool if sanity UI had some kind of `auto` radius property for cards 🤔

stipsan: For sure!

PR #1932, opened 2024-10-16
chore: dnd-ui-updates

georgedoescode:
This PR:

- Refactors DragInsertMarker to use inline styles vs styled components. SC was killing framerate in Safari during minimap sequences
- Adds an internal activeDragSequence variable to the overlay controller. This is currently used to prevent click events from firing shortly after drag sequences end — this occasionally caused the studio side to focus at an awkward time, resulting in frame-rate drops and unwanted scrollIntoView triggers.
- Adds onSequenceStart and onSequenceEnd callbacks to our handleDragAndDrop function, these are currently used to support the above, but could later be used to focus an item's parent array on drag start/end.
- Always renders skeleton rects. We used to only render them for drag previews of a certain size, but since we remove the drag handle it is no longer necessary.
- Uses a simple lerp to pick the best fitting radius value from the sanity UI design system for overlay drag previews
- Fixes preview rect stacking for elements with background images — we now always render image rects first, and text rects second.
- Adds proper drag cursor indication 
- Adds a marker around the active drag group 
- Ensures drag sequence listeners are cleaned up correctly


stipsan (approved):
Fantastic! 💖

PR #1946, opened 2024-10-17
chore(drag-and-drop): scroll-into-view, group bounds indicator fix

georgedoescode:
prevent scrollIntoView from firing while minimap open, fix group bounds overflow


stipsan (approved):
Thank you 💖

PR #1947, opened 2024-10-17
fix(presentation): observe correct preview snapshot perspectives

rdunk:
This should fix missing document labels on overlay elements corresponding to documents without draft versions when in the draft perspective.


stipsan (approved):
Splendid 💖

PR #1948, opened 2024-10-17
fix(visual-editing): always report document paths if unresolved

rdunk:
Fixes an issue where union types were not being resolved correctly as the ref used to compare which paths have been resolved was set _before_ a response was received from Presentation. Now we more aggressively query for union types, meaning the context menu for array items should no longer show "Unknown type".


georgedoescode (approved):
LGTM!

stipsan (approved):
Good catch! 💖

PR #1949, opened 2024-10-17
feat(visual-editing): allow mounting custom overlay components

rdunk:
This adds support for mounting custom React components inside of overlays. The `useDocuments` hook has been reworked to provide a more straightforward user facing API, and the `useOptimistic` hook now returns published (`id`) and original `(originalId)` IDs in the reducer action to avoid the need for comparing against/casting to a draft ID in userland.


stipsan (approved):
LGTM 💖

PR #1979, opened 2024-10-18
feat(visual-editing): add duplicate context menu action

rdunk:
Adds a duplicate action to the context menu for array items. Also refactors to use `doc.patch` where applicable.


stipsan (approved):
Good stuff! Let's go!

PR #1983, opened 2024-10-18
chore: enough with prettier and eslint fighting

stipsan:
I'm sick of [this](https://github.com/sanity-io/visual-editing/pull/1982) [fight](https://github.com/sanity-io/visual-editing/pull/1981)


georgedoescode (approved):
Nice! I think it makes sense to remove from eslint. 🚀

PR #1984, opened 2024-10-18
feat: lower permissions req to Collaborator

stipsan:
We used to require Editor or above to allow someone to use Presentation Tool when `previewUrl.previewMode.enable` was used, as it creates documents in your dataset with secrets, and used to have a `sanity-preview-url-secret.**` document `_id` prefix.
The `sanity-preview-url-secret.` prefix ensures the documents can't be queried without being Viewer or above. But the downside is that Collaborator roles can only create `drafts.` prefixes.
By moving to `drafts.` Collaborators can now use Presentation Tool 🥳 

Enterprise customers can with custom roles setup access for Viewer roles as well. We always supported this, but didn't document how to set it up. I've updated the README with a step-by-step for how to use it.

Viewers and below also now see a toast message when they can't create preview secrets:

image
Instead of a crash like we had before.

Applications that don't make use of `previewUrl.previewMode.enable` works just fine with Viewer and above on all price plans.

### DIFF
--- a/packages/next-loader/src/defineLive.tsx
+++ b/packages/next-loader/src/defineLive.tsx
@@ -97,7 +97,7 @@ export interface DefineSanityLiveOptions {
    */
   client: SanityClient
   /**
-   * Optional. If provided then the token needs to have permissions to query documents with `drafts.` prefixes as well as `sanity-preview-url-secret.` prefixes.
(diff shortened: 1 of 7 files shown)


EoinFalconer (approved):

evenwestvang:
This is great and in-line with policy. Thanks you for the attention to detail on this.

PR #1994, opened 2024-10-21
feat: add preview link sharing with QR code

stipsan:
# Introducing Preview Sharing :sparkles:

https://github.com/user-attachments/assets/ebadcbd1-a082-40b3-9814-04cd4dccb8e5

Preview Sharing creates share links with codes that are valid for as long as you want them to be. Toggling sharing off/on will create a new code that makes previous URLs no longer valid. 
It also shows a QR code to enable quickly opening a PR on your phone to test your content in context before publishing it.

## Prerequisites

For the menu to show up you'll need to have preview mode configured:

```tsx
presentationTool({
  previewUrl: {
    previewMode: { // or `previewMode`
      enable: '/api/draft-mode/enable' // <- required
    }
  }
})
```

And to use `@sanity/preview-kit` v2 in your app for calling `validatePreviewUrl` in your API route, v1 doesn't support the new custom-duration preview URL secrets used to authenticate the request to see draft content.

### Permissions

By default only Editor's or above can toggle sharing, although Enterprise plans with custom roles can change this.

## Opting out

The menu is enabled by default when `previewMode.enable` is setup. You can turn it off by setting `shareAccess` to `false`.
```tsx
presentationTool({
  previewUrl: {
    previewMode: {
      enable: '/api/draft-mode/enable',
      shareAccess: false // <- disable the QR code preview sharing menu
    }
  }
})
```


rdunk (approved):
Awesome, this is very cool.

PR #2002, opened 2024-10-21
fix(schema-extract): handle `title` sometimes being a React node

stipsan:
This one is a [bit tricky to reproduce on this repo](https://github.com/sanity-io/visual-editing/pull/2002/commits/8968474aaa1395caece33a03f6553631f4dd50ec), since it's also breaking `sanity schema extract` which we use for TypeGen:

```
      name: 'media',
      // @ts-expect-error this is actually allowed at runtime
      title: <>Media</>,
```
Having JSX in a `title` anywhere in the schema will cause a crash that causes the comlink connection to be lost:

Uncaught error

It's currently happening if you go to the [Test Studio](https://test-studio.sanity.dev/presentation/presentation/simpleBlock/eed1de44-4731-48f5-ab77-ea50a82dfef9?preview=%2Fpreview%2Findex.html) over at the `sanity` monorepo.

### DIFF
--- a/packages/@repo/sanity-schema/src/page-builder-demo/index.tsx
+++ b/packages/@repo/sanity-schema/src/page-builder-demo/index.tsx
@@ -30,7 +30,9 @@ const productType = defineType({
     defineField({
       type: 'array',
       name: 'media',
-      title: 'Media',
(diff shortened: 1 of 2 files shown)


rdunk (approved):
Thank you!

PR #2026, opened 2024-10-22
fix(visual-editing): resolve inline schema types when getting fields

rdunk:
This adds support for resolving inline fields, for example given the following schema types:

```ts
const fooType = defineType({
  type: 'object',
  name: 'foo',
  fields: [{type: 'string', name: 'title'}],
})

const barType = defineType({
  type: 'object',
  name: 'bar',
  fields: [
    {
      name: 'foo',
      type: 'foo',
    },
  ],
})
```

Visual editing will currently error when hovering over overlays which have a path that corresponds to some inline field (or a field nested inside one). This PR fixes that.


stipsan (approved):
Splendid 💖

PR #2035, opened 2024-10-23
chore(drag-and-drop): minimap stability, drag enablement

georgedoescode:
- Fixes a bug where certain `overflow` and `height` values on the document root allow scrolling while the minimap is open 
- Only enables drag and drop where `data-sanity` attribution is explicitly applied. Right now we have some issues mutating primitive arrays, and it feels generally safer (and in line with the docs) to require this.


stipsan (approved):
Thank you! 💖

PR #2036, opened 2024-10-23
fix: resolve issues with newly added canary features

rdunk:
This PR contains quite a few fixes related to features recently merged into `main` from `canary`. LMK if you'd like these to be split into separate PRs! I'd recommend reviewing each commit in turn if not.

Key commits explained below:

#### fix(visual-editing): support schema inline array items
Fixes a bug reported earlier today where trying to resolve an inline array item field from the schema would throw an error. High importance as hovering over overlay elements now triggers field resolution, so if an overlay's path points to an inline array item, an uncaught error is thrown.

#### fix: correctly handle popups
Previously new comlink targets for popups were added by using the `.size` of a Set stored as state as a `useEffect` dependency, but `.add()` was being called directly, rather than the `Set` being replaced using `setState`, which I believe failed to trigger the `useEffect`.

#### fix(presentation): suppress postMessage warnings
Prevents Presentation's comlink controller from initialising until the iframe has loaded, which suppresses the annoying `postMessage` warnings.

#### fix: support document snapshots on delayed comlink init
Fixes an issue where the welcome event was only emitted when Presentation initialised, and so would not be replayed to new window contexts, meaning document snapshots would not be fetched.

#### fix: enable optimistic features only when supported
Disables new features when communicating with an outdated version of Presentation that doesn't support mutations or in contexts where the optimistic actor is unavailable. Drag and drop, context menu, custom component mounting, etc. 

#### fix: explicit perspective fetching / fix: explicit schema fetching
Adds methods for explicitly fetching information needed when a context is initialised. Previously window contexts relied on Presentation blindly emitting some initial events, which means that any newly added contexts (popups) would not receive them.

### DIFF
--- a/packages/presentation/src/PostMessagePerspective.tsx
+++ b/packages/presentation/src/PostMessagePerspective.tsx
@@ -1,0 +1,31 @@
+import type {ClientPerspective} from '@sanity/client'
+import {memo, useEffect, type FC} from 'react'
+import type {VisualEditingConnection} from './types'
+
+export interface PostMessagePerspectiveProps {
+  comlink: VisualEditingConnection
+  perspective: ClientPerspective
+}
+
+const PostMessagePerspective: FC<PostMessagePerspectiveProps> = (props) => {
+  const {comlink, perspective} = props
+
+  // Return the perspective when requested
+  useEffect(() => {
+    return comlink.on('visual-editing/fetch-perspective', () => ({
+      perspective,
+    }))
+  }, [comlink, perspective])
+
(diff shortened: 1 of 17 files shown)


stipsan (dismissed):
Stellar work! Once the build is green and the merge conflicts are taken care of I say let's go 🏎️ 💨

rdunk:
> Stellar work! Once the build is green and the merge conflicts are taken care of I say let's go 🏎️ 💨

Damn, it's late. All resolved, and thanks for reviewing. 🙇

stipsan (approved):

PR #2055, opened 2024-10-25
chore(drag-and-drop): enable re-ordering of primitive array items

georgedoescode:


apps/page-builder-demo/src/sanity.types.ts
georgedoescode, line 1: should we git ignore this file? 🤔

stipsan: Hmm, I'd prefer that we report the unstable ordering to CLX and have it resolved instead 🤔

stipsan (approved):

apps/page-builder-demo/src/sanity.types.ts
georgedoescode: Nice, I'll raise it!

PR #2058, opened 2024-10-25
fix(visual-editing): remove useOptimistic nullable check

rdunk:
There are likely scenarios where you would want to use `useOptimistic` with a nullable passthrough value. For example when rendering a page with an optionally empty value.

This PR just removes the nullable `lastPassthrough` check. I'm sure it was there for good reason, but I haven't been able to find a situation where removing it causes issues so far.

@stipsan assigning you to this as you wrote it originally and probably have a good idea of why that check exists.


stipsan (approved):
Pretty sure the check was there because I wasn't aware of the nullable case, so I'm approving 👍

PR #2070, opened 2024-10-30
chore: add knip

rdunk:
[Knip](https://knip.dev/) is pretty cool, I know this was considered a while back, and there were reports of too many false positives from other devs that had used this internally, but it looks to be generally pretty good on this repo. So opening this as a nice to have in case others are interested.


stipsan (approved):
Awesome!

PR #2071, opened 2024-10-30
feat(visual-editing): allow passing props to overlay components

rdunk:
This PR:
- Adds the ability to pass props alongside components returned by the overlay component resolver.
- Adds a new `UnionInsertMenuOverlay` component export (via a new export path) that can be returned by the resolver for union types. This uses `InsertMenu` internally.
- Uses `InsertMenu` within the context of the `ContextMenu` if `insertMenu` config exists in the schema type.

### DIFF
--- a/packages/visual-editing/src/overlay-components/components/UnionInsertMenuOverlay.tsx
+++ b/packages/visual-editing/src/overlay-components/components/UnionInsertMenuOverlay.tsx
@@ -2,7 +2,14 @@ import type {SchemaNode, SchemaUnionNode} from '@repo/visual-editing-helpers'
 import {AddIcon} from '@sanity/icons'
 import type {SchemaType} from '@sanity/types'
 import {Button, Flex} from '@sanity/ui'
-import {useCallback, useRef, useState, type FunctionComponent, type MouseEvent} from 'react'
+import {
+  useCallback,
+  useRef,
+  useState,
+  type FunctionComponent,
+  type HTMLProps,
+  type MouseEvent,
+} from 'react'
 import styled from 'styled-components'
 import type {ElementNode, OverlayComponent} from '../../types'
 import {useDocuments} from '../../ui/optimistic-state/useDocuments'
@@ -11,7 +18,6 @@ import {InsertMenu} from './InsertMenu'
 
 const AddButton = styled(Button)`
   position: relative;
-
   transform: var(--add-button-position);
 
   --add-button-position: translateY(0);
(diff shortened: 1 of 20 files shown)


packages/visual-editing/src/overlay-components/components/UnionInsertMenuOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const [showButton, setShowButton] = useState(false)
+  const onEnter = useCallback(() => {
+    setShowButton(true)
+  }, [])
+  const onLeave = useCallback(() => {
+    setShowButton(false)
+  }, [])
+  const ref = useRef<HTMLDivElement | null>(null)
+
+  const bubbleEvent = useCallback(
```
georgedoescode, line 71: I think `bubble` may be slightly confusing here as the event isn't bubbling in the traditional sense. `transfer` or `delegate` or something cuold be good

packages/visual-editing/src/overlay-components/components/UnionInsertMenuOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+          bubbles: true,
+          cancelable: true,
+        })
+        element.dispatchEvent(newEvent)
+      }
+    },
+    [element],
+  )
+
+  const [referenceElement, setReferenceElement] = useState<HTMLElement | null>(null)
```
georgedoescode, line 85: Could be a little more explicit to point to the + button

packages/visual-editing/src/ui/context-menu/ContextMenu.tsx
```diff
@@ -55,22 +59,45 @@ function ContextMenuItem(props: {node: ContextMenuNode; onDismiss?: () => void})
         fontSize={1}
         icon={node.icon}
         padding={2}
+        // menu={{
```
georgedoescode, line 62: Can probably remove the commented menu prop?

packages/visual-editing/src/ui/context-menu/contextMenuItems.tsx
```diff
@@ -16,23 +16,23 @@ import {
...
-import {get} from '@sanity/util/paths'
+import type {SchemaType} from '@sanity/types'
+import {MenuGroup} from '@sanity/ui'
+import {type FunctionComponent} from 'react'
+import {InsertMenu} from '../../overlay-components/components/InsertMenu'
 import type {ContextMenuNode, OverlayElementField, OverlayElementParent} from '../../types'
 import {getNodeIcon} from '../../util/getNodeIcon'
-import {getArrayItemKeyAndParentPath} from '../../util/mutations'
-import {randomKey} from '../../util/randomKey'
+import {
```
georgedoescode, line 25: I like that we are moving the mutations out to live on their own 🙏

georgedoescode (approved):
A few tiny things but overall looks good! Thanks for the additional context over zoom :)

packages/visual-editing/src/ui/context-menu/ContextMenu.tsx
rdunk: Good point, I'll add a comment because this will be needed when a `@sanity/ui` PR lands.

packages/visual-editing/src/overlay-components/components/UnionInsertMenuOverlay.tsx
```diff
@@ -0,0 +1,192 @@
+import type {SchemaNode, SchemaUnionNode} from '@repo/visual-editing-helpers'
```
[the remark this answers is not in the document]
rdunk: Yes! I think it would be sensible to prefix this export path with `unstable_` for now though.

georgedoescode (approved):
Changes look good!

PR #2072, opened 2024-10-31
feat: releases v0.5

pedrobonamin:
### Pre read
This branch builds on top of `releases` branch, which includes the necessary packages updates.
- @sanity/client: `6.22.2-bundle-perspective`
- sanity: `corel` 
- @sanity/ui: local tar file (will be updated soon to a npm version)
- @sanity/icons:  local tar file (will be updated soon to a npm version)

### Next steps:
- [ ] Support for previews outside of the studio, think on adding the `bundlePerspective` data as a query param
- [ ] update `applySourceDocuments` in the `client` to work with perspectives. Investigate what this does.
- [ ]  Test things and find bugs.
- [ ] Creating a version document from presentation doesn't work as expected

## Changes:
With the introduction of `releases` the perspective is now a global studio concern and a new type of perspective has been introduced `bundle.{releaseId}`

**Example of a published release selection**
Screenshot 2024-10-31 at 13 58 23
Screenshot 2024-10-31 at 13 58 38

## Video 🎥 
Video explaining the details https://www.loom.com/share/3894e17c19b247548dccf3976f661a29?sid=59625a39-bdbf-400f-ad9c-d4243acf4018

The issue with the drafts is now fixed.

### Perspective selector removed 

Given we have this new perspective selector, the `draft | published` selector from Presentation tool [was removed ](https://github.com/sanity-io/visual-editing/compare/releases...releases-v0.5?expand=1#diff-a9073c81f31f1f1ed609e81ba6fec833f4ef8c482355c28162299318d6e6f7c2L410-L505)

### `<RevisionSwitcher/>` removed

This is not needed anymore, structure now knows about perspectives and it infers them from the global state and the form reacts according to the perspective selected.

### Loaders updated.

The loaders `LiveQueries` and `LoaderQueries` were updated to use the bundlePerspective param.

### Perspective type updated to include `bundle.${string}`

The global perspective can be any of `published | previewDrafts | bundle.${string}` this has been reflected in the code by updating the types were needed, this could be centralized instead of manually adding it everywhere, but wanted to easily reflect the intention in code for the review. 
This is relevant at the moment we need to query the data. When receiving a `bundle.${string}` perspective it should not use the `perspective` option in `client.fetch` , instead, it should use the `bundlePerspective` option, which takes an array of perspectives.

How does `bundlePerspective` works:
bundles perspective allows the client to fetch data in a similar way as `previewDrafts`, but instead of only stacking changes of published and drafts, it allows us to stack multiple perspectives inside.
So given you have the following documents:
```
-  foo, drafts.foo, versions.summer.foo
-  bar, versions.winter.bar
```
- If you use bundlePerspective="summer", you will get the documents `versions.summer.foo` and `bar`
- If you use bundlePerspective="winter,summer,drafts", you will get the documents `versions.summer.foo`and - -`versions.winter.bar` (This supports multiple perspectives in one query)
- if you use bundlePerspective="drafts", you will get `drafts.foo` and `bar`

This `bundlePerspective` array is exposed by the `usePerspective` hook from `sanity` which takes care of applying the perspectives in the correct order, the perspectives will apply on top of each other from right to left.
So if the same document is in two perspectives, it will be returned only in the first one found.


pedrobonamin:
Closed in favour of https://github.com/sanity-io/visual-editing/pull/2081

pedrobonamin:
Closed in favor of https://github.com/sanity-io/visual-editing/pull/2081

PR #2073, opened 2024-10-31
chore: set release scripts for corel

pedrobonamin:
adds release scripts for `corel`
Using `corel` tag as it matches with the tag we have in `sanity`


stipsan (approved):
Gogogo 🏎️ 💨

PR #2078, opened 2024-11-03
fix: backwards compatible handshake payloads

rdunk:
- Handshake messages in the internal `@sanity/channels` package sent the connection ID as the value of `data.id`.
- Comlink handshakes don't, and instead just check against the `connectionId` property.
- `data.id` isn't currently set in the comlink to channel compatibility transform.
- So handshakes currently fail when a newer version of `@sanity/presentation` is trying to communicate with an older version of `@sanity/visual-editing`.
- This PR adds the necessary `data.id` value in the compatibility transform.


georgedoescode (approved):
Assuming the rest of the message shapes are compatible, and the `id` vs `connectionId` formatting is the only issue here — LGTM!

stipsan (approved):

PR #2084, opened 2024-11-04
chore(release-overlay-previews): add simple release overlay previews

georgedoescode:


packages/presentation/src/PostMessageReleases.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {getVersionFromId, useClient} from 'sanity'
+import type {VisualEditingConnection} from './types'
+
+export interface PostMessageReleasesProps {
+  comlink: VisualEditingConnection
+  perspective: ClientPerspective | `bundle.${string}`
+  bundlesPerspective: string[]
+}
+
+const PostMessageReleases: FC<PostMessageReleasesProps> = (props) => {
```
pedrobonamin, line 12: This component will be used only inside the studio right? In `PresentationTool` 

Maybe we could reuse here the releases from `import {useReleases} from 'sanity'` , what do you think?

pedrobonamin: in that case, the studio will be responsible for providing the list of releases, which will change a bit in the near future. And you won't need to do this here. 

If it's inside the context of the studio the hook should work

packages/visual-editing/src/ui/ElementOverlay.tsx
```diff
@@ -299,15 +327,117 @@ export const ElementOverlay = memo(function ElementOverlay(props: ElementOverlay
...
+          (a, b) =>
+            new Date(a.metadata.intendedPublishAt).getTime() -
+            new Date(b.metadata.intendedPublishAt).getTime(),
+        )
+
+        nearestUpcomingRelease = scheduledReleasesSorted[0]
+
+        timeTilNextRelease = getTimeDifferenceInDays(
+          nearestUpcomingRelease.metadata.intendedPublishAt,
+        )
```
pedrobonamin, line 376: for sorting we also need to consider the release.publishAt date, which will be added once the release is scheduled. 
So 
```
   const scheduledReleasesSorted = scheduledReleases.sort(
          (a, b) =>
            new Date(a.publishAt || a.metadata.intendedPublishAt).getTime() -
            new Date(b.publishAt || b.metadata.intendedPublishAt).getTime(),
        )
```

And we could also have releases that are neither asap or scheduled, which are the `undecided`

Maybe it's easier for now to copy this from `sanity` to use the same sort mechanism. https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/core/releases/hooks/utils.ts#L6-L49

pedrobonamin: Also, if you are using releases from ` useReleases` you can trust that they will be sorted, so you could use the first match found.

packages/presentation/src/PostMessageReleases.tsx
georgedoescode: Ah awesome — this works! I initially used a simple query as eventually we would like to try this out in custom overlay components and it was good research, but we don't need it here. Have updated :)

packages/visual-editing/src/ui/ElementOverlay.tsx
georgedoescode: I am using `useReleases` now so all sorting logic is removed. Great call out!

packages/presentation/src/PostMessageReleases.tsx
```diff
@@ -0,0 +1,33 @@
...
+      comlink.post({
+        type: 'presentation/releases',
+        data: {
+          releases: releases.data,
+        },
+      })
+    }
+
+    run()
+  }, [comlink, perspective, bundlesPerspective])
```
pedrobonamin, line 28: You will probably want here `releases.data` as a dependency ?

georgedoescode: oops sorry good catch 👍

pedrobonamin (approved):
Thanks for this @georgedoescode ! It is amazing to see the releases in the overlays.

PR #2114, opened 2024-11-07
fix(docs): update img tags to markdown format

markmichon:
We experienced some parsing issues when rendering this file on the docs site. Unless there are specific stylistic needs, it's preferable to use markdown's built-in image syntax rather than html in readme files for better compatibility.


stipsan (approved):
Thanks!

PR #2117, opened 2024-11-08
fix(visual-editing): correct component resolver array return types

rdunk:
Small change to the overlay component resolver return type.

Moves the union to the array itself rather than applying the union at array level, and so allowing for arrays of mixed types.


georgedoescode (approved):
Nice! 🚀

PR #2118, opened 2024-11-08
feat(presentation): allow passing custom header component

rdunk:
PR to allow passing a custom preview header component via `components.unstable_header` in the plugin config, similar to `unstable_navigator`.

This component passed replaces the entire preview header bar, so it also receives a `renderDefault` prop for convenience. In practice, this allows users to prepend or append a button/control/whatever to the preview header, as in this screenshot. This is maybe not fine-grained enough, as users may want to insert a component at some intermediate position.

Screenshot

--- 

Example component which appends a button:
```tsx
import type {PreviewHeaderProps} from "@sanity/presentation"
// other imports...

export const CustomHeader: FunctionComponent<PreviewHeaderProps> = (props) => {
  return (
    <>
      {props.renderDefault(props)}
      <Button
        icon={HomeIcon}
        onClick={() => {...}}
      />
    </>
  );
};
```

Please be honest with feedback/ideas for changes, this is not urgent so may be good to discuss more before merging.


stipsan (approved):
LGTM 💖

PR #2163, opened 2024-11-14
fix(comlink): ensure nodes always accept handshake synchronize events

rdunk:
Currently, comlink's node state machine will only handle syn handshake events when in an `idle` state. In some situations, the controller or some of the channels it maintains may detect a loss of connection and transition to a `handshaking` state, the disconnect event emitted at this point may then fail to reach the relevant node(s), and thus the node will remain in a `connected` state, as the nodes themselves only blindly respond to heartbeat events.

This PR adds a change so that the node state machine will _always_ transition to its internal `handshaking` state when receiving a syn handshake event, regardless of the state it is currently in. This should greatly improve the success rate of attempt reconnects. For example you may have noticed in our test environments that the connection is often lost after a HMR, this should no longer be the case.

It also drops the default expected response timeout to 3s from 10s for requests for all requests that expect a response except for those initiated with `.fetch` (which remains at 10s).

### DIFF
--- a/packages/comlink/src/controller.ts
+++ b/packages/comlink/src/controller.ts
@@ -260,8 +260,7 @@ export const createController = (input: {targetOrigin: string}): Controller => {
 
     const stop = () => {
       channels.forEach((channel) => {
-        channel.disconnect()
-        channel.stop()
+        cleanupChannel(channel as unknown as Channel<Message, Message>)
       })
     }
 
--- a/packages/comlink/src/constants.ts
+++ b/packages/comlink/src/constants.ts
@@ -4,7 +4,10 @@ import type {MessageType} from './types'
 export const DOMAIN = 'sanity/comlink'
 
 /** @internal */
-export const RESPONSE_TIMEOUT = 10000
+export const RESPONSE_TIMEOUT_DEFAULT = 3_000
+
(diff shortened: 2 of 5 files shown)


packages/comlink/src/controller.ts
```diff
@@ -260,8 +260,7 @@ export const createController = (input: {targetOrigin: string}): Controller => {
 
     const stop = () => {
       channels.forEach((channel) => {
-        channel.disconnect()
-        channel.stop()
+        cleanupChannel(channel as unknown as Channel<Message, Message>)
```
azaxarov, line 263: Will the following also work?
```suggestion
        cleanupChannel(channel)
```

stipsan (approved):
Great catch! :O 💖

packages/comlink/src/controller.ts
rdunk: Unfortunately not as it stands, I guess this has to do with type variance? `channel` here is of type `Channel<R, S>`, and that isn't considered a subtype of `Channel<Message, Message>` even if `R` and `S` extend `Message`...or something like that.

I need to look at improving the typing in places, if you have any suggestions they'd be very welcome! 😊

PR #2180, opened 2024-11-18
feat(visual-editing): add async getSnapshot to optimistic documents

rdunk:
This is kind of a fix dressed up as a feat.

Accessing optimistic document snapshots is currently error prone as it is assumed the snapshot will be available, in actual fact it may not be as snapshots have to be fetched from Content Lake via Presentation tool. In most situations this is fine as user interaction is required to trigger some document snapshot getter and so will _likely_ not occur immediately

`doc.get()` returns a snapshot and throws if the snapshot is unavailable. This is now deprecated.

`doc.getSnapshot()` replaces it. It returns a promise which is resolved when the snapshot is actually returned.

As a result the Context Menu now has a loading state (video below demonstrates this with an artificial 5s delay added to snapshot fetching), and interactions like drag and drop will await the snapshot resolution before patches are applied, rather than just throwing.

Similarly `doc.patch((context) => {...}` will currently throw if used before snapshots have been fetched. In this PR `snapshot` is deprecated but still available on this context object, however the property has been replaced with a getter so that it will not throw unless explicitly accessed. `snapshot` is replaced with `getSnapshot` which similarly returns a promise.

https://github.com/user-attachments/assets/38ad7c1b-68b8-4d17-95f9-da2ee5523ff3

### DIFF
--- a/apps/page-builder-demo/src/components/overlays/ExcitingTitleControl.tsx
+++ b/apps/page-builder-demo/src/components/overlays/ExcitingTitleControl.tsx
@@ -1,5 +1,3 @@
-'use client'
-
 import {at, set} from '@sanity/mutate'
 import {get} from '@sanity/util/paths'
 import {SanityNode, useDocuments, type OverlayComponent} from '@sanity/visual-editing'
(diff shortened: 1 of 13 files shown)


stipsan (approved):
Makes sense to me!

Issue #2179, opened 2024-11-18
BUG: Overlay cursors remain when the edit is disabled


mariusGundersen:
When hovering over elements the cursor is set to either `move` or `auto` from the code here: https://github.com/sanity-io/visual-editing/blob/10ecc2a791876501bb55aa92956d6c177f5b2a50/packages/visual-editing/src/controller.ts#L151

Unfortunately this is not cleaned up when edit is toggled up, which means the cursors look different when trying to navigate the page normally.

For example, we have set all `a:href` to have `cursor:pointer`, but with the overlay they get the default pointer. This means that when the overlay is disabled it looks like something is wrong, since they don't have the pointer they normally have. 

Worse yet is that for array items that can be rearranged they will have the `move` cursor even though they cannot be moved.

https://github.com/user-attachments/assets/a354fa37-c1a7-4c25-81a9-c6fa72637fa2

rdunk:
Thanks for reporting! I've opened a PR with a potential fix https://github.com/sanity-io/visual-editing/pull/2181.

rdunk:
#### Fixed in
https://github.com/sanity-io/visual-editing/pull/2181
#### Released in
https://github.com/sanity-io/visual-editing/releases/tag/presentation-v1.18.5
https://github.com/sanity-io/sanity/releases/tag/v3.65.0

PR #2181, opened 2024-11-18
fix(visual-editing): unset overlay applied cursor on mouseleave

rdunk:
Fix for #2179. This unsets the cursor style when `mouseleave`-ing an overlay element.  Previously on `mouseleave` the element was specified as the `hoveredElement`, which I think would be the element that the cursor was entering after leaving the one we want to reset?

Two questions:
1. Should we only set cursor to `auto` on child elements of drag groups?
2. As this manipulates the DOM outside of our visual editing element, should we store and revert to any user set cursor style value?

@georgedoescode I've assigned this to you as you'll know the original intention!

### DIFF
--- a/packages/visual-editing/src/controller.ts
+++ b/packages/visual-editing/src/controller.ts
@@ -127,7 +127,17 @@ export function createOverlayController({
     })
   }
 
-  function setOverlayCursor() {
+  function setOverlayCursor(element: ElementNode, remove?: boolean) {
+    if (remove) {
+      handler({
+        type: 'overlay/setCursor',
+        element,
+        cursor: undefined,
+      })
+
+      return
+    }
+
     if (!inFrame || !optimisticActorReady) return
 
     const hoveredElement = getHoveredElement()
@@ -149,7 +159,7 @@ export function createOverlayController({
 
     handler({
       type: 'overlay/setCursor',
-      element: hoveredElement,
+      element,
       cursor,
     })
   }
@@ -272,7 +282,7 @@ export function createOverlayController({
           rect: getRect(element),
         })
 
-        setOverlayCursor()
+        setOverlayCursor(element)
       },
       mouseleave(e) {
         function leave() {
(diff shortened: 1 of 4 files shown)


rdunk:
Pushed a new commit which implements both the suggestions in the original description.

packages/visual-editing/src/controller.ts
```diff
@@ -127,30 +129,56 @@ export function createOverlayController({
...
+
+      // Resolve the drag group for the currently looped over stack element
+      const dragGroup = resolveDragAndDropGroup(el, targetSanityData, elementSet, elementsMap)
+
+      if (dragGroup) {
+        // If the element in the stack is the currently hovered element, set the
+        // cursor to 'move'. Otherwise the element is a child ancestor, so set
+        // it to 'auto' to override the parent's move cursor
+        const isHoveredElement = element === el
+        cursor = isHoveredElement ? 'move' : 'auto'
```
georgedoescode, line 153: I'm wondering — do we need to modify the children's cursor values? I feel like if we don't need to show the drag handle, we can just do nothing?

packages/visual-editing/src/ui/Overlays.tsx
```diff
@@ -136,7 +136,11 @@ const OverlaysController: FunctionComponent<{
       } else if (message.type === 'overlay/setCursor') {
         const {element, cursor} = message
 
-        element.style.cursor = cursor
+        if (cursor) {
+          element.style.cursor = cursor
+        } else {
+          element.style.removeProperty('cursor')
```
georgedoescode, line 142: Related to the above: if we don't modify children's `cursor` values, we may not need this `removeProperty`?

packages/visual-editing/src/controller.ts
rdunk: Huh, this is a good point actually... This workaround of checking the hover stack is necessary because our deferred leave pattern means the leave handler isn't triggered until the element is removed from the hover stack, so the parent element maintains its cursor style and the child elements need to use `auto` to override the cascade.

It might be possible to just not set the cursor style in the deferred leave handler, but in the mouseleave event itself. That would simplify things.

georgedoescode (approved):
This is working really well. Nice one!

PR #2194, opened 2024-11-25
fix: prevent disconnect on window context close

rdunk:
To address this issue: https://github.com/sanity-io/next-sanity/issues/1950

Uses a `setInterval` to detect when opened window contexts have closed, so that we don't try to maintain connections to closed popups.

Also includes a small comlink improvement to correctly cleanup unused connections.

### DIFF
--- a/packages/presentation/src/PresentationTool.tsx
+++ b/packages/presentation/src/PresentationTool.tsx
@@ -66,6 +66,7 @@ import type {
 import {useDocumentsOnPage} from './useDocumentsOnPage'
 import {useMainDocument} from './useMainDocument'
 import {useParams} from './useParams'
+import {usePopups} from './usePopups'
 import {usePreviewUrl} from './usePreviewUrl'
 import {useStatus} from './useStatus'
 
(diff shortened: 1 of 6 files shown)


packages/presentation/src/useStatus.ts
```diff
@@ -1,35 +1,59 @@
...
 export function useStatus(): [string, (event: StatusEvent) => void] {
+  // State to keep track of the status of each channel
   const [statusMap, setStatusMap] = useState(
     new Map<string, {status: Status; hasConnected: boolean}>(),
   )
 
+  // Memoized computation of the overall status based on the status of individual channels
   const status = useMemo(() => {
     const values = Array.from(statusMap.values())
+    // If any channel is connected, return the `connected` status
```
rdunk, line 28: This was moved up so that the status is always "connected" if at least one channel is currently connected. This is to prevent the Preview component from re-rendering each time a popup is created.

packages/presentation/src/usePopups.ts
```diff
@@ -0,0 +1,83 @@
...
+          unsubs.push(controller.addTarget(source))
+        }
+      }
+    }
+    return () => {
+      unsubs.forEach((unsub) => unsub())
+    }
+  }, [controller, popups])
+
+  // Checks for closed popups and removes them from the tracked Set
```
rdunk, line 53: Most of this is just existing logic moved to a hook.

However this useEffect is new. It checks to see if existing popups have been closed at a given interval (1000ms).

This is to prevent comlink attempting to reconnect to a non-existent Node when the heartbeat detects a disconnect.

packages/presentation/src/PresentationTool.tsx
rdunk, line 1: This is just refactoring, popup related logic has moved to `usePopups.ts`

stipsan (approved):

Issue #2165, opened 2024-11-14
Expose createCompatibilityActors function and the PreviewKitNodeMsg type


osnoser1:
**Is your feature request related to a problem? Please describe.**

I've finished and deployed the visual-editing and preview-kit feature in Angular; with the exposure of the channels package into `@sanity/comlink`, I was able to remove it from the repo; however, for implementing these features, I made an angular equivalent of [useDocumentInUse.ts](https://github.com/sanity-io/visual-editing/blob/main/packages/preview-kit-compat/src/useDocumentsInUse.ts) here [document-in-use.service.ts](https://github.com/limitless-angular/limitless-angular/blob/main/packages/sanity/preview-kit-compat/src/document-in-use.service.ts), and I used the two new dependencies `createCompatibilityActors` function and the `PreviewKitNodeMsg` type of the `@repo/visual-editing-helpers`, I copied the project partially.

**Describe the solution you'd like**

It should be great to expose these two things through a npm package.

rdunk:
Just wanted to provide you with an update as we've discussed this internally: we're planning on restructuring this monorepo and moving away from unpublished internal packages, so we should be able to solve this one soon. Sorry it's taken a while, thanks for your patience and the amazing work you're doing to support Angular.

PR #2207, opened 2024-11-26
feat(comlink)!: public api improvements

rdunk:
This PR introduces some changes to the public API of `@sanity/comlink` that I've wanted to make for a while. There are no new features, this is mainly refactoring, but there are breaking changes. All consumers of the package in this repo have been updated accordingly, including the compatibility actors.

This isn't strictly necessary from our point of view, but as this package is likely going to be used by other teams, I felt the improvements would be useful.

### Terminology

In the initial version, I think some concepts were incorrectly named. The concepts of `Channels` and `Connections` have now been swapped, which I think makes much more sense:

(Apologies for the unorthodox use of diff block here 😄 )
```diff
-- Controllers create Connections which maintain Channels to Nodes

++ Controllers create Channels which maintain Connections to Nodes.
```

### `post` and `fetch` signatures

The signatures for `.post()` and `.fetch()` methods have changed:
```diff
-- comlink.post({type: 'visual-editing/toggle', data: {enabled: true}})

++ comlink.post('visual-editing/toggle', {enabled: true})
```

This is more concise and aligns with `.on()`, which already looks like this:
```ts
comlink.on('visual-editing/toggle', (data) => {})
```

### Type argument order

The order of type arguments has been updated in many interfaces and functions. Placing `Sends` before `Receives` feels more natural and aligns better with expectations:

```diff
-- export interface ChannelInstance<R extends Message, S extends Message> {}

++ export interface ChannelInstance<S extends Message, R extends Message> {}
```


stipsan (approved):
Stellar work @rdunk! Absolutely love this 💖 
As far as I'm concerned we should merge and publish this asap 🏎️ 💨 
Looks like the new API signature has better typing and autocomplete as well :chefs-kiss:
A nice follow up could be to move the compatibility actor out of `@repo/visual-editing-helpers` and to `@sanity/comlink`, perhaps on its own export path, to resolve https://github.com/sanity-io/visual-editing/issues/2165 💖

rdunk:
> A nice follow up could be to move the compatibility actor out of `@repo/visual-editing-helpers` and to `@sanity/comlink`, perhaps on its own export path, to resolve #2165 💖

That's a great idea! Thanks.

PR #2235, opened 2024-11-28
fix: use CSS view transition for viewport iframe toggling

stipsan:
Whenever the preview is showing a layout that isn't super simple the viewport transition is quite janky and jarring:

https://github.com/user-attachments/assets/c11f4d32-7b9c-4199-89f8-2cef21658f4f

By moving it to CSS View Transitions instead of having Framer Motion animate the `<iframe>` `width` and `height` it's much smoother:

https://github.com/user-attachments/assets/def0e72b-360a-4649-b65f-fe374c196ff2

### DIFF
--- a/apps/page-builder-demo/src/sanity.types.ts
+++ b/apps/page-builder-demo/src/sanity.types.ts
@@ -771,6 +771,104 @@ export type ProjectsPageQueryResult = Array<{
   slug?: Slug
 }>
 
+// Source: ./src/app/product/[slug]/page.tsx
+// Variable: productSlugsQuery
+// Query: *[_type == "product" && defined(slug.current)]{"slug": slug.current}
+export type ProductSlugsQueryResult = Array<{
+  slug: string | null
+}>
+// Variable: productPageQuery
+// Query: *[_type == "product" && slug.current == $slug][0]
+export type ProductPageQueryResult = {
+  _id: string
+  _type: 'product'
+  _createdAt: string
+  _updatedAt: string
+  _rev: string
+  title?: string
+  slug?: Slug
+  media?: Array<{
+    asset?: {
+      _ref: string
+      _type: 'reference'
+      _weak?: boolean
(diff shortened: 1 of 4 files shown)


rdunk:
This is awesome, but do we lose animations in Safari + FF with this change?

Just tested very quickly, I assumed the residual framer motion stuff was intended as a fallback, but I don't see that on those browsers.

stipsan:
> This is awesome, but do we lose animations in Safari + FF with this change?
> 
> Just tested very quickly, I assumed the residual framer motion stuff was intended as a fallback, but I don't see that on those browsers.

Great callout @rdunk! I was contemplating falling back to framer, which is why I commented out the variant instead of deleting it 🤔  I'll add it for Ff! What version of Safari did you test btw? [Caniuse says we should be good to go](https://caniuse.com/mdn-api_document_startviewtransition) and it worked on my test run 🤔 🤔 🤔

stipsan:
I was using Safari Technical Preview! 🤦 🤦 🤦

rdunk:
Ah yeah, I'm still on Safari 17, looks like 18 supports which is GA, so maybe we're good there. FF, [maybe soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1909173)?

FWIW ditching framer motion for this I think is fine, the fallback to just not having an animation seems graceful enough to me, you still get the intended feature, but I'll leave it up to you!

stipsan:
> Ah yeah, I'm still on Safari 17, looks like it's supported in 18 which is GA, so maybe we're good there. FF, [maybe soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1909173)?
> 
> FWIW ditching framer motion for this I think is fine, the fallback to just not having an animation seems graceful enough to me, you still get the intended feature, but I'll leave it up to you!

I think I fixed it, can you try now @rdunk? 🤗

stipsan:
Merged as I tested on Firefox, Safari 17, Safari TP and it works ☺️

PR #2240, opened 2024-12-02
feat(next-loader): add option to disable stega in defineLive

rdunk:
Small fix to adjust how stega is set in `SanityLiveStream` and `sanityFetch`.

Currently, the value passed as a prop/param is used if set, otherwise `stega` is set to true if:
- Draft mode is enabled.

With this PR, the value passed as a prop/param is used if set, otherwise `stega` is set to true if:
- Draft mode is enabled.
- The client config has a defined `stega.studioUrl`.
- `stega` has not been explicitly set to `false` when calling `defineLive`.

### DIFF
--- a/packages/next-loader/src/defineLive.tsx
+++ b/packages/next-loader/src/defineLive.tsx
@@ -112,6 +112,11 @@ export interface DefineSanityLiveOptions {
      */
     revalidate?: number | false
   }
+  /**
+   * Optional. Include stega encoding when draft mode is enabled.
+   *  @defaultValue `true`
+   */
+  stega?: boolean
 }
 
 // export type VerifyPreviewSecretType = (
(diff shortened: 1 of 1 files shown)


packages/next-loader/src/defineLive.tsx
```diff
@@ -179,7 +179,7 @@ export function defineLive(config: DefineSanityLiveOptions): {
     perspective?: Omit<ClientPerspective, 'raw'>
     tag?: string
   }) {
-    const stega = _stega ?? (await draftMode()).isEnabled
+    const stega = _stega ?? ((await draftMode()).isEnabled && client.config().stega)
```
stipsan, line 182: I'm not sure this works.

We want to allow this pattern:
```
const client = createClient({
  // ...
  stega: {
    studioUrl: '/studio' <- stega is ready, but not enabled
  }
})

export const {sanityFetch, SanityLive} = defineLive({client})
```

We don't want stega to be enabled in production, but if userland has to do this:
```
const client = createClient({
  // ...
  stega: {
    enabled: true, <- applies to both production and draft mode
    studioUrl: '/studio'
  }
})

export const {sanityFetch, SanityLive} = defineLive({client})
```

stipsan (approved):
Thank you! ❤

PR #2262, opened 2024-12-04
fix(comlink): require data param if defined on post and fetch methods

rdunk:
Small fix for comlink that improves the types for `post` and `fetch` methods.

Previously, I added functionality to make the second/data param optional so that:
```ts
// Given a message type without data
type MsgWithoutData = {
  type: 'some-message-without-data'
  data: undefined
}
// You can call 
node.post('some-message-without-data')
// Instead of having to always explicitly pass two params
node.post('some-message-without-data', undefined)
```
But this has the unintended side effect that:
```ts
// Given a message type with data
type MsgWithData = {
  type: 'some-message-with-data'
  data: { foo: string }
}
// You could call this without an error
node.post('some-message-with-data')
```

These changes mean the second parameter is now required if the type's `data` param is not undefined
```ts
// So this will now error
node.post('some-message-with-data')
// And you have to do this
node.post('some-message-with-data', { foo: 'bar' })
```

### DIFF
--- a/packages/comlink/src/connection.ts
+++ b/packages/comlink/src/connection.ts
@@ -62,7 +62,9 @@ export type Connection<S extends Message = Message, R extends Message = Message>
     handler: (event: U['data']) => Promise<U['response']> | U['response'],
   ) => () => void
   onStatus: (handler: (status: Status) => void) => () => void
-  post: <T extends S['type'], U extends Extract<S, {type: T}>>(type: T, data?: U['data']) => void
+  post: <T extends S['type'], U extends Extract<S, {type: T}>>(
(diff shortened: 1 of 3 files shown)


georgedoescode (approved):
Pulled locally and can confirm, for the event `visual-editing/toggle`: 

- `comlink?.post('visual-editing/toggle')` throws as expected 
- `comlink?.post('visual-editing/toggle', {})` throws as expected 
- `comlink?.post('visual-editing/toggle', {foo: '1234'})` throws as expected 
- `comlink?.post('visual-editing/toggle', {enabled: true})` works as expected. 

LGTM!

PR #2263, opened 2024-12-04
chore: add visual-editing telemetry

georgedoescode:
This PR adds a `sendVisualEditingTelemetry` function that leverages comlink to send telemetry events using Studio's `useTelemetry` hook.


packages/visual-editing/src/ui/telemetry/sendTelemetry.ts
```diff
@@ -0,0 +1,34 @@
...
+  }),
+}
+
+type EventDataMap = {
+  [K in keyof typeof events]: (typeof events)[K] extends ReturnType<typeof defineEvent<infer T>>
+    ? T
+    : never
+}
+
+function sendTelemetry<K extends keyof typeof events>(
```
georgedoescode, line 18: I debated doing this a bunch of ways, but as this is an early iteration I feel like a simple function is a safe bet. We can expand as we need?

apps/studio/.env.local
```diff
@@ -12,6 +12,8 @@ SANITY_STUDIO_PAGE_BUILDER_DEMO_PREVIEW_URL=""
 SANITY_STUDIO_REMIX_PREVIEW_URL=""
 SANITY_STUDIO_SVELTE_PREVIEW_URL=""
 
+SANITY_STUDIO_PRESENTATION_DEBUG_TELEMETRY=true
```
stipsan, line 15: You'll need to also add this to `apps/studio/turbo.json` I think

georgedoescode: Oh cool! It works without but want to be sure. Have added :)

packages/presentation/src/PostMessageTelemetry.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const PostMessageDocumentVersions: FC<PostMessageTelemetryProps> = (props) => {
+  const {comlink} = props
+
+  const telemetry = useTelemetry()
+
+  useEffect(() => {
+    const isDev = process.env['NODE_ENV'] === 'development'
+    const debugTelemetry = process.env['SANITY_STUDIO_PRESENTATION_DEBUG_TELEMETRY'] === 'true'
```
stipsan, line 16: The way it's setup here would mean that the code we publish on npm has this:
```
process.env['SANITY_STUDIO_PRESENTATION_DEBUG_TELEMETRY'] === 'true'
```
It's a bit tricky, as we have people who embed their sanity studios on envs that do not have `process` available as a global.
And for certain systems, like next.js, it wants you to use this format:
```
process.env.SANITY_STUDIO_PRESENTATION_DEBUG_TELEMETRY === 'true'
```
Which our ts linting generally don't like.


To make this fully safe to use, regardless of bundlers and runtimes that will consume this library, you'd need to do something like:
```
const debugTelemetry = 
  (typeof process !== 'undefined' &&
   typeof process.env !== 'undefined') ?
process.env.SANITY_STUDIO_PRESENTATION_DEBUG_TELEMETRY === 'true' :
   false
```

Or better yet, we could use `debug` like they do in the studio 🤔

stipsan (commented):
How can you disable telemetry?

georgedoescode:
> How can you disable telemetry?

Hello! I know we spoke about this offline but adding here for consistency: 

Thanks to this PR https://github.com/sanity-io/sanity/pull/7958 we can now leverage the same `SANITY_STUDIO_DEBUG_TELEMETRY` as Studio, and automatically have telemetry be a noop with logging in the browser. 

`SANITY_STUDIO_DEBUG_TELEMETRY` is set in our `.env.local` for `apps/studio`. 

Chatting with Bjoerge, I know that the studio does not prevent telemetry for local studio instances, so I don't think we need to worry about any of that, and all the bespoke `process.env` logic has been removed!

packages/presentation/src/PostMessageTelemetry.tsx
```diff
@@ -0,0 +1,36 @@
...
+  comlink: VisualEditingConnection
+}
+
+const PostMessageDocumentVersions: FC<PostMessageTelemetryProps> = (props) => {
+  const {comlink} = props
+
+  const telemetry = useTelemetry()
+
+  useEffect(() => {
+    const isDev = process.env['NODE_ENV'] === 'development'
```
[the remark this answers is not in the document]
georgedoescode: Resolving as we don't need this!

packages/presentation/src/PostMessageTelemetry.tsx
georgedoescode: We are now entirely leaning on Studio's telemetry implementation, and this looks to be handled here: https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/studio/StudioTelemetryProvider.tsx#L30

packages/visual-editing/src/ui/telemetry/sendTelemetry.ts
stipsan: Yep I think it's fine 👍

stipsan (dismissed):
💖

packages/visual-editing/src/ui/telemetry/sendTelemetry.ts
```diff
@@ -0,0 +1,34 @@
+import {defineEvent} from '@sanity/telemetry'
+import type {VisualEditingNode} from '../../types'
+
+const events = {
+  'Visual Editing Drag Sequence Complete': defineEvent({
+    name: 'Visual Editing Drag Sequence Complete',
+    description: 'An array is successfully reordered using drag and drop.',
+    version: 1,
+  }),
+}
```
georgedoescode, line 10: @svirs is this event formatting OK?

packages/visual-editing/src/ui/telemetry/sendTelemetry.ts
georgedoescode: Excellent :)

packages/visual-editing/src/ui/telemetry/sendTelemetry.ts
svirs: @georgedoescode I'd make the event name past-tense `Completed`

georgedoescode: Thanks!

stipsan (approved):

PR #2307, opened 2024-12-10
feat(visual-editing): add svelte optimistic state support

rdunk:
This PR adds optimistic state support for Svelte via the `useOptimistic` export in `@sanity/visual-editing/svelte`.

I've also re-worked a lot of the existing optimistic related exports, and added a new `@sanity/visual-editing/optimistic` export path which exposes the non-framework specific parts needed for e.g. Nuxt to implement this functionality.

Going forward we should recommend that the React version of `useOptimistic` be imported from `@sanity/visual-editing/react` rather than the root `@sanity/visual-editing` export path. I've added that export to the React specific path but haven't removed anything from that root path for now, so no breaking changes. I'm not sure how we can mark the root exports specifically as deprecated with TSDoc, open to ideas there!

So the two big things to review are:
1. Do the changes to exports make sense?
2. Does the Svelte `useOptimistic` implementation follow the existing React logic OK?


stipsan (approved):
I've refactored the deprecation markings to follow the same approach we use for marking:
```ts
import createClient from '@sanity/client' // tsdoc says you should migrate to `import {createClient} from '@sanity/client'`
```
in `@sanity/client`.

>Do the changes to exports make sense?

Yep!


>Does the Svelte useOptimistic implementation follow the existing React logic OK?

And yep! Let's go!

PR #2308, opened 2024-12-10
fix(visual-editing): execute fetches after node connect event

rdunk:
This is an attempted fix for this potential race condition we're seeing where sometimes features that rely on the optimistic actor aren't working as expected.

Instead of nodes fetching (via comlink) the data we need to initialise optimistic features once on app load, we instead perform these fetches each time the node status changes to `connected`. The hope is that if some disconnect happens, we fetch this data when we have recovered, rather than assuming these fetches will always succeed.

N.b. I haven't had time to test this one as thoroughly as I'd have liked.


stipsan (approved):
Thank you! We'll give it a thorough try 🙇 💖

PR #2356, opened 2024-12-25
fix(visual-editing): Don't show "Open in Studio" offscreen when at top of screen

DAlperin:
For things like navbars or logos that might appear on the top of the screen, the overlay could be cut off. This moves the "Open in Studio" below the element in that case.

Supplants #1868 

Before:
image
After:
image


stipsan (approved):
Thank you! 💖

PR #2363, opened 2025-01-03
feat: add optional redirect override to svelte preview handler

rdunk:
The Svelte specific preview mode functionality we've implemented in the visual editing package imports `redirect` from `@sveltejs/kit`. However if the consuming app is using mismatched dependency versions the redirect will fail. See https://github.com/sveltejs/kit/issues/11749#issuecomment-1966565367 for more details as to why.

The PR just adds the option to pass a `redirect` function explicitly to override that imported by the package itself, which should ensure that preview mode redirects to the correct location instead of throwing an error.


stipsan (approved):
Nice! 💖

rdunk:
Rebased to fix merge conflicts.

Also added `@sanity/visual-editing` to peer deps of `@sanity/svelte-loader`.

stipsan (approved):

PR #2474, opened 2025-01-09
feat(comlink)!: update internal message types, buffer unhandled messages

rdunk:
This PR introduces unhandled message buffering, so that if some event handler is added after a related event is received, the last n (defaults to 1) events can be replayed. This should resolve the race condition issue we've been seeing.

It also introduces a change I've wanted to make for a while, specifically to remove user defined events being emitted directly from the `node` and `connection` actors. All Comlink messages are now emitted via the `'message'` event type, so the state machine will no longer directly emit messages like `'visual-editing/focus'`. This means we no longer need to prefix state machine related events (referred to as "internal events") with an underscore to differentiate them from user defined events, and it also helps improves typings.

As these internal event types are accessible via the `onInternalEvent` method, this is a breaking change.

I've also refactored generic type names to follow convention as they were unnecessarily ambiguous, i.e. `S` -> `TSends`, `R` -> `TReceives`.

### DIFF
--- a/packages/comlink/playground/app/routes/_index.tsx
+++ b/packages/comlink/playground/app/routes/_index.tsx
@@ -43,17 +43,15 @@ export default function Index() {
 
     setChannel(channel)
 
-    channel.onInternalEvent('_message', (event) => {
-      console.log('_message', event)
-      setReceived((prev) => [event.message, ...prev])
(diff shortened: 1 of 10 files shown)


stipsan (approved):
Brilliant, let's merge and release 🥳🎉

PR #2486, opened 2025-01-13
feat(presentation): listen for changes on draft document

nkgentile:
Adjust document store listener to subscribe to changes on both the draft and published documents. Currently, while listener receives draft changes using the `previewDrafts` perspective, it doesn't receive events emitted while changes are being made to the draft.


stipsan (approved):
Tested, works well!! 💖

PR #2487, opened 2025-01-13
fix: explicitly fetch preview snapshots on connect

rdunk:
Ensure the visual editing package explicitly requests preview snapshots for Presentation on connect, this should fix the issue we're seeing with Astro.

Currently snapshots are only returned from Presentation when the preview snapshot observable emits. This is a problem if for example the app inside the iframe reloads, as the observable will not emit until some change is made.


georgedoescode (approved):
LGTM!

PR #2556, opened 2025-01-23
chore: remove `@sanity/preview-kit-compat`

stipsan:
This package is no longer needed, as `@sanity/preview-kit` since v5.2.0 no longer uses it. It now registers as a loader, similar to `@sanity/next-loader`, in order to support `corel` https://github.com/sanity-io/preview-kit/pull/1513


rdunk (approved):
Nice! That was an easy one to review 😄.

PR #2581, opened 2025-01-28
fix: remove vX override

stipsan:
It's no longer necessary to use vX with Draft Live Content 🥳


rdunk (approved):
LGTM, I'll do the same for the new Svelte package.

PR #2597, opened 2025-01-30
fix: increase heartbeat interval from 1s to 10s

stipsan:
Increasing the interval reduces the gc work Presentation has to do when keeping track of queries. Especially when moving back and forth between URLs.


rdunk (approved):
LGTM.

PR #2615, opened 2025-02-05
fix(visual-editing): prevent snapshot overfetching

rdunk:
It looks like https://github.com/sanity-io/visual-editing/commit/74fdcddacb7ef02748cf729b67c6cdbb4c3647e2 introduced some changes in the `DocumentReporter` component which removed the unique document ID check.

This is causing documents to be unobserved/observed on every render—such as hovering on/off overlay—even if the IDs haven't changed, which in turn triggers a request to Presentation to fetch new document snapshots for every document on the page. This can result in hundreds of requests on each interaction. See video below.

This PR reverts that change, and should fix this issue.

https://github.com/user-attachments/assets/71fe6447-99b7-4ee9-91b9-bf62152b0583

### DIFF
--- a/packages/visual-editing/src/ui/Overlays.tsx
+++ b/packages/visual-editing/src/ui/Overlays.tsx
@@ -78,26 +78,33 @@ const DocumentReporter: FunctionComponent<{
   perspective: ClientPerspective
 }> = (props) => {
   const {documentIds} = props
+  const [uniqueIds, setUniqueIds] = useState<string[]>([])
+
+  useEffect(() => {
+    setUniqueIds((prev) => {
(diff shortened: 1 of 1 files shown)


packages/visual-editing/src/ui/Overlays.tsx
```diff
@@ -323,6 +330,7 @@ export const Overlays: FunctionComponent<{
   }, [overlayEnabled])
 
   const documentIds = useMemo(() => {
+    console.log('recalc documentIds')
```
stipsan, line 333:
```suggestion

```

stipsan (approved):
Great catch!! 😱

PR #2636, opened 2025-02-06
fix: solve race in comlink.onStatus

stipsan:
I've tried to screencap the before/after of this one, but it's notoriously difficult. The bug manifests itself as `status` in `<Overlays />` being stuck as `idle`, thus overlay elements never render: https://github.com/sanity-io/visual-editing/blob/d0702d8b5264d287bfcc751bbb6e02a553567fb2/packages/visual-editing/src/ui/Overlays.tsx#L347

With the latest stable of `@sanity/visual-editing` I can reproduce it after 10 or so attempts of rapidly switching between structure tool and presentation tool.
With `@sanity/visual-editing@2.12.15-canary.0` I've attempted to reproduce almost 100 times and it no longer happens 😮‍💨

### DIFF
--- a/apps/page-builder-demo/src/sanity.types.ts
+++ b/apps/page-builder-demo/src/sanity.types.ts
@@ -871,24 +871,6 @@ export type PageSlugsResult = Array<{
   slug: string | null
 }>
 
-// Source: ./src/app/project/[slug]/page.tsx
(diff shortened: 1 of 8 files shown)


rdunk (approved):
Cool, LGTM, we can improve how comlink handles status later.

PR #2993, opened 2025-05-12
refactor(apps): use @sanity/sveltekit in svelte app

rdunk:
Gentle PR to ease back in... This refactors the Svelte app to use the new `@sanity/sveltekit` package which supersedes the exports at `@sanity/visual-editing/svelte` and `@sanity/svelte-loader`. We should be able to deprecate those packages shortly, when docs and guides have been updated. I'll also publish a first major of `@sanity/sveltekit`.

It also updates the app to use Svelte 5 syntax, making use of new features like runes.


stipsan (dismissed):
Let's go! ❤

PR #2996, opened 2025-05-19
feat: Custom Overlay Plugins

jmswrnr:
This PR introduces a higher-level plugins API for custom overlays in the presentation tool.

This work can be split into 2 main focus areas:
- The new `data-sanity-edit-group` - groups multiple sanity nodes into one visual overlay element.
- The new `plugins` prop - a simple pattern to add custom overlay components.

### Questions
- [x] Should we call this Plugins instead of Extensions?
- [ ] We have the new `OverlayPluginDefinition` type, but should we also provide a public type for plugin definitions in function form to encourage a nice pattern which accepts user-configurable plugin options.

### Considerations
- [ ] `next-sanity` will need to update [`<VisualEditing />`](https://github.com/sanity-io/next-sanity/blob/a6f893b2c6a55337b63b62a57ddc1eb7400c3014/packages/next-sanity/src/visual-editing/client-component/VisualEditing.tsx#L134) component to pass this new `plugins` prop.

### Review Focus Points
- `findSanityNodes` has been refactored to support groups of multiple nodes to one visual overlay element. The returned `ResolvedElement` data is expected to have a `sanity` node, but now with the introduction of groups can contain multiple targets (including the `sanity` node). I've kept a `commonSanity` value in this `ResolvedElement` which is the same `sanity` node for individual overlay elements with one sanity node, but for groups containing multiple sanity nodes it attempts to find a common node and otherwise falls back to the first node if no common path is found. This fallback isn't ideal but reduces changes in this PR by keeping a guaranteed sanity node per visual overlay element, even if it just falls back to the first. We could improve this as follow up work.
- The `controller.tsx` has been updated to support groups of multiple sanity nodes connected to one visual overlay element. This work includes some strategic unregistering of element and re-parsing of groups on mutation. Would be great to get a sense check on this part, check as when groups are added/removed and elements inside a group are mutated and should still remain inside a group and not register again.
- `<PointerEvents />` is now used in `ElementOverlay.tsx` - I'm not sure if this should be used here or not, there seems to be a strange occurrence to do with the hover stacking pattern causing an overlay to get stuck open when an exclusive plugin is closed without the element being hovered anymore.


packages/visual-editing/src/ui/ElementOverlay.tsx
```diff
@@ -189,59 +240,194 @@ const ElementOverlayInner: FunctionComponent<ElementOverlayProps> = (props) => {
...
+            </Labels>
+          </Tab>
+        )}
+
+        <HUD>
+          {plugins?.map((plugin, i) =>
+            plugin.hud.map((hud) => {
+              const Component = hud.component
+              if (!Component) return null
+              return <Component key={i} PointerEvents={PointerEvents} {...plugin.context} />
```
rdunk, line 416: Here `i` is the index of the outer `plugins?.map`, so is the same key is being assigned to all components from the inner `hud.map`?

packages/visual-editing/src/ui/ElementOverlay.tsx
```diff
@@ -43,7 +52,10 @@ const isReactElementOverlayComponent = (
 }
 
 export interface ElementOverlayProps {
+  comlink?: VisualEditingNode
+  comlinkStatus?: Status
```
rdunk, line 56: Is the `comlinkStatus` prop used by the `ElementOverlay`?

rdunk:
Also dropping a note here that we'll also need to check about adding any new props for other frameworks too once this gets merged.

packages/visual-editing/src/ui/ElementOverlay.tsx
jmswrnr: I tidied this up but there still seems to be a duplicate key error coming from somewhere, will investigate further

packages/visual-editing/src/ui/ElementOverlay.tsx
jmswrnr: Nope, removed this prop now

packages/visual-editing/src/ui/ElementOverlay.tsx
jmswrnr: Ah ok `context.node.id` is not a unique ID, switched that out for a runtime `uuid`

stipsan (dismissed):
Looking great! Stellar work 💖

jmswrnr:
Thanks @stipsan ! 🙌 

Would be great to resolve this PointerEvents pattern here before bringing this in, @rdunk - there's a [new commit](https://github.com/sanity-io/visual-editing/pull/2996/commits/967e8834d4d77b309fc08dd4606b6c02d7dd2207) to tackle this, could you please check that out! it's a little hacky but seems to be working as expected. When closing an exclusive plugin control, it should now allow you to click outside the overlay plugin area to close and reset the hover stack, then if you're hovering an overlay element after it closes, that new overlay element should be visible.

jmswrnr:
Thanks for the feedback @rdunk! 🙌 

> I wonder if the ellipsis menu should follow the same behaviour established by the other popover interactions

I've now locked body scroll using the popover background, and allow the mouse to leave without closing the over which can now click-to-edit. 8d6fb31

> I'm not sure about the first menu item.

This was added because granular click-to-edit is lost when grouping multiple sanity nodes to the one visual element. Perhaps we could hide this label for non-group overlays, but I left it in for consistency and the fact it needs to label what the plugin is running on (can have same plugin option multiple times on one menu for different nodes).

Screenshot 2025-05-23 at 10 54 35

> The use of `<PointerEvents>` to wrap the content of the `<ElementOverlayInner>` means the label tab will now always receive events

I've now updated this label so clicking will now focus like clicking the main element. 5c595d3

> Should the label tab stay active when an exclusive plugin is enabled?

I don't think so, when making an exclusive plugin the visual can overflow the element bounds with controls, for example a hotspot tool wouldn't want a label on top of the crop handle. Same with the close button, which is why I added the prop to close the plugin so users can implement a close method best suited for the plugin. As it's the only thing displayed when active, it's fine to let users implement custom methods, perhaps we could follow up with an opt-in close button component if there is need for it.

Also, the label isn't overly useful in group contexts also, as it picks the first sanity node if no common path is found which may not be related to the node you're editing in exclusive mode.

rdunk (approved):
Nice work, thanks for considering the feedback, let's get this merged finally! 🚀

PR #2997, opened 2025-05-19
fix(comlink): Promise.withResolvers shim

rdunk:
Adds a fallback implementation of `Promise.withResolvers` used in the `fetch` method on nodes. Error was reported for dashboard users on older versions of Safari.

### DIFF
--- a/packages/comlink/src/node.ts
+++ b/packages/comlink/src/node.ts
@@ -35,6 +35,7 @@ import type {
   StatusEmitEvent,
   WithoutResponse,
 } from './types'
+import {createPromiseWithResolvers} from './util'
 
 /**
  * @public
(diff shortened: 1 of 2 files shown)


ChrisNolan:
Is there an opportunity here to have a system in place that knows what the minimum version of the main browsers are for the different javascript function used by the system and just simply fail gracefully telling users they need to upgrade their browser?

In this case it looks like there is a simple enough shim to get around it, but I'd think there would be other cases where this function was being used in other parts of studio which haven't been caught here?  Or other javascript features/improvements?

Just wondering out loud.  

E.g. to know the versions of browsers which have support for this function see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers

EoinFalconer (approved):

PR #3026, opened 2025-05-28
feat: add overlay plugin define helper

jmswrnr:
This PR:
- Adds a `defineOverlayPlugin` helper function (following #2996)
- Refactors existing demo app overlay components into new plugin API

The new `defineOverlayPlugin` function provides a pattern for defining typed overlay plugins with some features:
- A plugin can define a `guard` function to show which contexts it can support (for example all `string` fields). With the helper, it adds a user-configurable `guard` function to specify which project-specific contexts to use the plugin on (for example all `title` paths).
- A plugin can now also define an object of user-configurable options.

### What to review
- Guard logic is working as intended, the plugins are only shown on the intended overlay elements. Both guard functions (user config and plugin) need to pass in order for the plugin to be shown. If either guard function is not defined it will act as if `true` (passing).
- Title & 3D Rotate examples working as expected.


packages/visual-editing/src/overlay-components/defineOverlayPlugin.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        if (pluginGuardResult === false) {
+          return false
+        }
+
+        return config?.guard?.(context) ?? true
+      },
+    }
+  }) as OverlayPluginUserFn<O>
+}
```
rdunk, line 62: Nice!

apps/page-builder-demo/src/components/overlay-plugins/index.tsx
```diff
@@ -1,14 +1,20 @@
...
+  LEDLifespanHUD(),
+  ImageResolutionHUD(),
+  Rotate3D({guard: ({node}) => node.path.endsWith('rotations')}),
+  ExcitingTitlePlugin({
+    guard: ({node}) => node.path.endsWith('title'),
+    options: {append: '?', buttonText: '🤩'},
+  }),
+  ExcitingTitlePlugin({
+    guard: ({node}) => node.path.endsWith('headline'),
+  }),
```
rdunk, line 19: This is great.

rdunk (approved):
Awesome addition.

Also, feel free to remove and/or change the "ExcitingTitlePlugin" to an exclusive plugin if it makes more sense in the context of this demo app and what you need it to do.

rdunk (approved):
LGTM!

PR #3172, opened 2025-07-30
feat(comlink): add pause method

joshuaellis:
### Description

Dashboard would _love_ a way to "pause" channels, but not delete and destroy them (which is what the `stop` method appears to do). Because our iframes are kept "alive" by the activity API. So instead this method just runs the "cleanup" function that disconnects & stops connections, but they're kept in the registries so the start function should work once more when called.

### Disclaimer

This PR is just a "how in my naive mind thinks it could work" – it'd be good to know, what tests to add (if applicable).

### DIFF
--- a/packages/comlink/src/controller.ts
+++ b/packages/comlink/src/controller.ts
@@ -33,6 +33,7 @@ export interface ChannelInstance<TSends extends Message, TReceives extends Messa
     ...params: (TMessage['data'] extends undefined ? [TType] : never) | [TType, TMessage['data']]
   ) => void
   start: () => () => void
+  pause: () => () => void
   stop: () => void
 }
 
@@ -278,12 +279,20 @@ export const createController = (input: {targetOrigin: string}): Controller => {
       return stop
     }
 
+    const pause = () => {
+      const connections = channel.connections as unknown as Set<Connection>
+      connections.forEach(cleanupConnection)
+
+      return start
+    }
+
     return {
       on,
       onInternalEvent,
       onStatus,
       post,
       start,
+      pause,
       stop,
     }
   }


gu-stav:
@rdunk Thanks for taking a look!

In Dashboard we are using the Activity API from React to keep iframes alive in the background and therefore we want to keep the connections/ controller alive, but put the channels on pause while they are in the background. Once an iframe comes back into the current, we'd like to start the channel again. Does that make sense? 

There is a POC implementation here https://github.com/sanity-io/ada/blob/16f5a02d93ec6e531ed55a4d894290a75b07e009/apps/ada/app/features/comlink/ComlinkContext.tsx#L128-L144

rdunk:
Hey @gu-stav. Got that much, thanks! I think I’m still a bit fuzzy on the specifics of the problem, in other words: what's the current app behaviour without this? Is the connection not resuming properly? Or is it more of a performance/resource thing?

gu-stav:
@rdunk At the moment dashboard only keeps one controller per origin and destroys it, when an iframe/ origin is unmounted (and creates a new controller, if the iframe is mounted again).

We want to change that, so that if an iframe unmounts dashboard keeps the controller and channels alive, because with the Activity API the iframe is still there, but hidden.

For the time the iframe is hidden, we want to pause the channels for two reasons: 1) to pause all events (e.g. because an iframe that is hidden shouldn't be able to trigger a navigation) and 2) to be able to know when the connection between an iframe and dashboard has been established again, once the iframe is restored (e.g. because on of the channels has gone into state `connected` again).

Does that make sense? I could also walk you through the problem and code we are trying to solve - maybe there is a better approach.

PR #3492, opened 2026-07-13
feat(visual-editing): strip stega from clipboard on copy, add `onSuspiciousStega` reporting

stipsan:
### Description

Part of SAPP-3204 (strip stega from paste into plain text fields). This is the app/preview side of the equation — a follow-up PR in `sanity-io/sanity` will handle paste handlers in the Studio itself, so the ticket stays open after this merges.

#### Strip stega from the clipboard on copy (default on)

Content on a preview page contains stega-encoded metadata (invisible characters), which tags along when people copy text and paste it into other tools — plain text fields, URLs, spreadsheets, etc.

While Visual Editing is enabled, `copy` events are now intercepted and stega is stripped from the clipboard:

- Both the `text/plain` and `text/html` flavors are cleaned, so rich-text pastes stay clean too, and stega hidden in attributes of the copied markup (e.g. `img[alt]`) is also removed.
- Selections inside `<input>`/`<textarea>` fields are handled via the field's own selection.
- Copies that don't contain any stega are left completely untouched (the regex pre-test is cheap), and userland `copy` handlers that call `event.preventDefault()` take precedence.
- `cut` events are deliberately not intercepted, since `preventDefault()` on cut would require reimplementing content deletion.

Opt out with the new `keepStegaOnCopy` option/prop:

```tsx
<VisualEditing keepStegaOnCopy />
```

#### Opt-in `onSuspiciousStega` reporting

New callback that reports stega payloads found in places where they always cause a bug or bloat. If the callback isn't provided, no scanning runs at all.

```tsx
<VisualEditing
  onSuspiciousStega={(reports) => {
    for (const report of reports) console.warn(`Stega found in ${report.kind}`, report)
  }}
/>
```

Reported placements (`report.kind`):

| `kind` | What it means |
| --- | --- |
| `attribute` | Any attribute except the expected stega carriers (`img[alt]`, `time[datetime]`, `svg[aria-label]`): `class` (selectors no longer match), `id` (broken anchors/`getElementById`), `href`/`src`/`srcset`/`action` and other URL attributes (invisible characters end up percent-encoded in requests), `style`, `name`, `value`, `data-*` (broken equality checks), etc. |
| `head` | Anywhere inside `<head>` (text or attributes), e.g. `data.title` rendered in `<title>`, `meta[content]`, JSON-LD — never rendered, pure bloat, corrupts SEO/social metadata |
| `script` | Text content of `<script>` in body (JSON-LD, embedded state) |
| `style` | Text content of `<style>` |
| `form-value` | `<textarea>` content that would be submitted along with user input |
| `url` | The page URL itself (tested raw and percent-decoded), meaning the page was reached through a link with stega baked in |

Each report includes the offending `element`, `attribute` name (when applicable), the raw `value`, the `cleaned` value it should have been, and the decoded `sanity` edit info when available — pointing at the exact document/field that produced the value.

Efficiency: an initial full audit (TreeWalker over `head` + `body` + `<html>` attributes) and a dedicated `MutationObserver` that only re-tests what changed (a single attribute for attribute mutations, only unsafe-context text nodes for text mutations, added subtrees for childList). All work is deferred to `requestIdleCallback` (with `setTimeout` fallback), reports are deduped by `kind + attribute + cleaned value` and batched per flush. Body text nodes outside `<script>`/`<style>`/`<textarea>` are never tested — stega in rendered text is the feature, not a bug. Everything inside `<sanity-visual-editing>` is ignored.

#### Wiring

Both options are added to `VisualEditingOptions` and forwarded through all variants: `enableVisualEditing`, the React `<VisualEditing />` (used by `next-sanity`'s App Router component via its dependency on this package), `next-pages-router`, `react-router` and the Svelte component. Callbacks are routed through `useEffectEvent` so unstable identities don't re-initialize visual editing or re-run the audit.

### What to review

- Naming and JSDoc of `keepStegaOnCopy` / `onSuspiciousStega` / `SuspiciousStegaReport` in `packages/visual-editing/src/types.ts`
- Copy interception logic in `packages/visual-editing/src/util/stegaCleanOnCopy.ts`
- The unsafe-placement map and scanning strategy in `packages/visual-editing/src/util/suspiciousStega.ts`

### Testing

- New Vitest coverage: `stegaCleanOnCopy.test.ts` (8 tests: both clipboard flavors, attribute stega in HTML flavor, input/textarea selections, no-stega/defaultPrevented/collapsed no-ops, listener disposal) and `suspiciousStega.test.ts` (6 tests: initial audit incl. allowlist and rendered-text exclusions, mutation-driven reports for added nodes/changed attributes/changed text, dedupe, overlay exclusion, disposal), plus a `stegaClean` unit test.
- `pnpm --filter @sanity/visual-editing test`, `lint` and `build` (includes `pkg check --strict`) pass.

<div><a href="https://cursor.com/agents/bc-9d034ff4-cb63-4c03-8e28-da1c50005f33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9d034ff4-cb63-4c03-8e28-da1c50005f33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/strip-stega-on-copy.md
+++ b/.changeset/strip-stega-on-copy.md
@@ -1,0 +1,9 @@
+---
+'@sanity/visual-editing': minor
+---
+
+feat: strip stega from clipboard on copy, and report suspicious stega placements
+
+While Visual Editing is enabled, `copy` events are now intercepted and stega-encoded metadata (invisible characters) is stripped from the clipboard, so content copied from a preview can be pasted into other tools without the hidden characters tagging along. Copies without stega are left untouched, and the behavior can be disabled with the new `keepStegaOnCopy` option.
+
+Also adds an opt-in `onSuspiciousStega` callback that reports stega payloads found in places where they always cause bugs or bloat, such as `class`, `href`, `src`, `id` and other attributes, inside `<head>` (e.g. `<title>`), `<script>` or `<style>` contents, form values, or the page URL. Reports are deduped, batched, and include the decoded edit info when available so the source field can be tracked down.
--- a/packages/visual-editing/README.md
+++ b/packages/visual-editing/README.md
@@ -447,14 +447,14 @@ Providing the callback opts in to the detection logic — when it isn't provided
 
 Reported placements (`report.kind`):
 
-| `kind` | What it means |
-| --- | --- |
-| `attribute` | Stega in an attribute such as `class` (selectors no longer match), `id` (broken anchors and `getElementById`), `href`/`src` and other URL attributes (the invisible characters end up percent-encoded in requests), `style`, `name`, `value` or `data-*` (broken equality checks). |
-| `head` | Stega anywhere inside `<head>`, e.g. rendering `data.title` in `<title>` or `meta[content]`. It's never visible, so it's pure bloat that also corrupts SEO and social metadata. |
-| `script` | Stega inside a `<script>` element, e.g. JSON-LD structured data or embedded state. |
-| `style` | Stega inside a `<style>` element, breaking selectors or values. |
-| `form-value` | Stega in a form field value (e.g. `<textarea>` content), where it would be submitted along with user input. |
-| `url` | Stega in the page URL itself, meaning the page was reached through a link that had stega encoded into it. |
+| `kind`       | What it means                                                                                                                                                                                                                                                                      |
+| ------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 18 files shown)


jordanl17 (approved):
Great stuff!

PR #3620, opened 2026-08-12
feat(loaders): accept any @sanity/client version via SanityClientLike

rexxars:
## The problem

Loader options that take a client were typed as `SanityClient | SanityStegaClient`. That breaks when the app resolves a different `@sanity/client` than the loaders do:

```
error TS2322: Type 'SanityClient' is not assignable to type 'SanityClient | SanityStegaClient'.
  Property '#private' in type 'SanityClient' refers to a different member
  that cannot be accessed from within type 'SanityClient'.
```

The cause is not the v7 to v8 changes. `SanityClient` declares a `#private` field, which makes it **nominal** rather than structural, so a client only satisfies it when it comes from the exact same copy of the package. A duplicate install of the *same* version fails too. The v7/v8 config diffs (`requester` moving into `InitializedClientConfig`, `resolveFetch` replacing `_requestHandler`) are just extra fuel.

The methods the loaders actually use are unchanged: `fetch`, `config` and `withConfig` are byte-identical between v7.26.2 and v8.0.0, which is why one interface can serve both.

## The change

`SanityClientLike` in `@sanity/visual-editing-types`, covering only what the loaders need, re-exported from `@sanity/core-loader` so the react and svelte loaders pick it up through their existing `export type * from`:

```ts
export interface SanityClientLike {
  config(): {
    dataset?: string
    perspective?: ClientPerspective
    projectId?: string
    stega: InitializedStegaConfig
    token?: string
    useCdn: boolean
  }
  withConfig(config: SanityClientLikeConfig): SanityClientLike
  fetch<R = any>(query: string, params?: QueryParams, options?: /* filtered */): Promise<R>
  fetch<R = any>(query: string, params: QueryParams, options: /* unfiltered */): Promise<{result: R; resultSourceMap?: ContentSourceMap}>
}
```

Notes on the shape:

- Query options are `Pick<ResponseQueryOptions, ...>` rather than hand-written fields, so they track the real client instead of drifting from it.
- `config().stega` stays `InitializedStegaConfig`, which is what lets `enableLiveMode` pass it straight into `stegaEncodeSourceMap`. Safe because `StegaConfig` is structurally identical across majors and carries no private brand.
- Both `fetch` overloads are needed: the loaders use `filterResponse: false`, and `preview-url-secret`'s `validatePreviewUrl` uses the plain form.
- `R` defaults to `any` to match `SanityClient['fetch']`. Defaulting to `unknown` is stricter than the client it stands in for and breaks callers who cannot name the type.

It lives in `visual-editing-types` because that is the only existing package reachable from both sides: `core-loader` and `visual-editing` already pull it in transitively via `visual-editing-csm`.

Eight `as SanityClient` casts became unnecessary and are gone.

## Not a breaking change

Every option accepts strictly more than before: `createQueryStore({client})`, `setServerClient()`, `enableLiveMode({client})`, `useLiveMode({client})`, `handlePreview({client})`, `handleLoadQuery({client})`.

For SvelteKit, `event.locals.client` keeps the full `SanityClient` API:

```ts
export interface VisualEditingLocals<TClient extends SanityClientLike = SanityClient> {
  client: TClient
  preview: boolean
}
```

The default keeps `.listen()`, `.getDocument()` and the rest available to everyone already reading it. Our own `global.d.ts` stubs parameterise it with `SanityClientLike`, which is what lets `handlePreview` assign internally without a cast. Apps that resolve their own copy of `@sanity/client` name it explicitly in `app.d.ts`:

```ts
declare global {
  namespace App {
    interface Locals extends LoaderLocals<SanityClient> {}
  }
}
```

The one narrowing is `unstable__serverClient.instance`, now `SanityClientLike`. Leaving it as the nominal class would have recreated the bug in a different spot, since it holds a client the caller handed us. It is marked `@internal` and prefixed `unstable__`.

## Tests

`packages/core-loader/test/SanityClientLike.test-d.ts` installs `@sanity/client@8` under an alias as a devDep so the cross-major case is genuinely reproduced. Without a second copy in the tree the assertions would be vacuous, since one copy is always compatible with itself. It covers:

- clients from both majors, plain and stega, satisfy the interface
- the interface is accepted by `validatePreviewUrl`, which `handlePreview` calls
- the interface has not been quietly narrowed back to the nominal class
- non-clients are still rejected

`packages/svelte-loader/test/locals.test-d.ts` covers the locals default and the escape hatch.

## Also in here: the typecheck was inert

`typecheck.tsconfig` pointed at `tsconfig.build.json` in both packages. In `core-loader` that sets `noCheck`; in `svelte-loader` it only includes `src`, so files under `test` were outside the program. Either way every type assertion silently passed, including the new ones. Confirmed by asserting `expectTypeOf<string>().toEqualTypeOf<number>()` and watching it pass. Both now point at `tsconfig.json`, and each was verified to fail on a deliberate bad assertion before being trusted.

That flushed out a real bug in this PR: with no default on `R`, it fell back to `unknown` and broke the internal cache typing in `core-loader/src/index.ts`. Hence the `= any` above.

**Still inert, not fixed here:** `react-loader`'s existing `test/createQueryStore/universal.test-d.ts` and all of `visual-editing`'s tests, for the same two reasons. Turning them on risks surfacing a pile of unrelated pre-existing errors, so it is left for a follow-up, but they currently guard nothing.

## Out of scope

- `visual-editing/src/optimistic/*` keeps `SanityClient`. It uses `transaction`, `mutate` and `listen` and wants the real thing.
- `preview-url-secret`'s own narrower `SanityClientLike` is untouched. Its need is genuinely smaller, and widening it would break anyone passing a hand-rolled object. A type test locks the one-way compatibility instead.
- `packages/visual-editing-standalone/package.json` has a one-line `@sanity/ui` 4.0.1 to 4.0.2 change that the build regenerated. Drift from the recent `@sanity/ui` bump, unrelated to this work. Happy to drop it.
- Noticed but not changed: `defineUseLiveMode` in both loaders does `(studioUrl ?? typeof client === 'object') ? client?.config().stega.studioUrl : undefined`, which ignores an explicitly passed `studioUrl` whenever it is truthy. Looks like it was meant to be `studioUrl ?? (typeof client === 'object' ? ... : undefined)`. Only the cast was removed.

### DIFF
--- a/.changeset/loose-clients-decoupled.md
+++ b/.changeset/loose-clients-decoupled.md
@@ -1,0 +1,46 @@
+---
+'@sanity/visual-editing-types': minor
+'@sanity/core-loader': minor
+'@sanity/react-loader': minor
+'@sanity/svelte-loader': minor
+'@sanity/visual-editing': minor
+---
+
+feat: accept any `@sanity/client` version in the loaders via `SanityClientLike`
+
+Options that took a client used to be typed as `SanityClient | SanityStegaClient`. `SanityClient`
+declares a `#private` field, which makes it nominal rather than structural, so a client only
+satisfied it when it came from the exact same copy of `@sanity/client`. Passing a client from a
+different major failed to typecheck, and so did a duplicate install of the same version:
+
+```
+Type 'SanityClient' is not assignable to type 'SanityClient | SanityStegaClient'.
+  Property '#private' in type 'SanityClient' refers to a different member that cannot be
+  accessed from within type 'SanityClient'.
+```
+
+These options now take `SanityClientLike`, a structural interface covering only what the loaders
+use: `config()`, `withConfig()` and `fetch()`. Any client satisfies it, from any version, including
+the ones from `@sanity/client/stega`, `@sanity/preview-kit/client` and `next-sanity`.
+
+This affects `createQueryStore({client})`, `setServerClient()`, `enableLiveMode({client})`,
+`useLiveMode({client})`, `handlePreview({client})` and `handleLoadQuery({client})`. All of them
+accept strictly more than before, so no changes are needed.
+
+For SvelteKit, `event.locals.client` still gives you the full `SanityClient` API. If your app
+resolves a different copy of `@sanity/client` than these packages do, name your own client type in
+`app.d.ts` to avoid a mismatch:
+
+```ts
+import type {SanityClient} from '@sanity/client'
+import type {LoaderLocals} from '@sanity/svelte-loader'
+
+declare global {
+  namespace App {
+    interface Locals extends LoaderLocals<SanityClient> {}
+  }
+}
+```
+
+The one narrowing is `unstable__serverClient.instance`, which is now `SanityClientLike`. It is
+marked `@internal` and prefixed `unstable__`.
--- a/.changeset/typecheck-not-inert.md
+++ b/.changeset/typecheck-not-inert.md
@@ -1,0 +1,11 @@
+---
+'@sanity/core-loader': patch
+'@sanity/svelte-loader': patch
+---
+
+fix: run vitest's typecheck against a tsconfig that actually checks the test files
+
+`typecheck.tsconfig` pointed at `tsconfig.build.json` in both packages. In `core-loader` that config
+sets `noCheck`, and in `svelte-loader` it only includes `src`, so type assertions in `test` were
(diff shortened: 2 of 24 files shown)


stipsan (dismissed):
Thanks!

rexxars:
Rebased and ready for a re-review.

stipsan (approved):

